PR #5, opened 2024-05-29
ex-command, stylua and q to close

mosheavni:
## Changes

- quit floats with <kbd>q</kbd>
- offer some completion to `:Kubectl` ex-command
- Open the existing views if we try `:Kubectl get {existing_view}`
  I'm not a big fan of the `if..elseif..elseif..else` so if you have a better alternative, please offer.
- fix stylua filename and indentation by Tabs (this is what the repo is using, I'd rather using spaces but it's your choice, just align stylua accordingly)

### DIFF
--- a/lua/kubectl/actions.lua
+++ b/lua/kubectl/actions.lua
@@ -25,6 +25,8 @@ function M.new_buffer(content, filetype, opts)
 
 	if opts.is_float then
 		layout.float_layout(buf, filetype, opts.title or "", opts.syntax or filetype)
+		vim.bo[buf].buflisted = false
+		vim.keymap.set("n", "q", vim.cmd.close, { buffer = buf, silent = true })
 	else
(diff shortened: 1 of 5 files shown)


Ramilito:
> * quit floats with q

Love it!
> * offer some completion to `:Kubectl` ex-command

Nice!
> * Open the existing views if we try `:Kubectl get {existing_view}`
>   I'm not a big fan of the `if..elseif..elseif..else` so if you have a better alternative, please offer.

I might have one but we can start with this.
> * fix stylua filename and indentation by Tabs (this is what the repo is using, I'd rather using spaces but it's your choice, just align stylua accordingly)

Lol nice catch, will format it to be spaces and lower the width in a separate PR

Ramilito:
Something like this for the if/else: https://github.com/Ramilito/kubectl.nvim/commit/6229be7175f2dde7143747a1f67c070165067502

Issue #23, opened 2024-06-01
feat: more views


Ramilito:
**Describe the solution you'd like**

A list of views we want:

- [x] CRDS
- [x] ClusterRoleBindings
- [x] ClusterRoles -- fallback view is enough
- [x] Roles --fallback view is enough
- [x] RoleBindings --fallback view is enough
- [x] CONFIGMAPS
- [x] STATEFULSETS -- fallback view is enough
- [x] DEAMONSET
- [x] PV
- [x] PVC
- [x] SA
- [x] INGRESS -- fallback view is enough
- [x] Fallback view

idelice:
Configmaps is done in PR #42 .

Ramilito:
crds added in #154 
daemonsets added in #153

Ramilito:
pv added in #155 
pvs added in #161 
sa added in #162

Ramilito:
crb added in #173

Ramilito:
With the awesome fallback view by @mosheavni, I think we can close this PR! Really great feature ❤️

PR #42, opened 2024-06-12
feat: view shortcuts + fix error callback

idelice:
Implemented Configmaps view as part of https://github.com/Ramilito/kubectl.nvim/issues/23#issue-2328787749

### DIFF
--- a/ftplugin/k8s_pods.lua
+++ b/ftplugin/k8s_pods.lua
@@ -1,11 +1,14 @@
 local api = vim.api
-local container_view = require("kubectl.views.containers")
-local deployment_view = require("kubectl.views.deployments")
 local hl = require("kubectl.actions.highlight")
 local loop = require("kubectl.utils.loop")
-local pod_view = require("kubectl.views.pods")
 local tables = require("kubectl.utils.tables")
 local view = require("kubectl.views")
+local configmaps_view = require("kubectl.views.configmaps")
+local container_view = require("kubectl.views.containers")
+local deployments_view = require("kubectl.views.deployments")
+local pod_view = require("kubectl.views.pods")
+local secrets_view = require("kubectl.views.secrets")
+local services_view = require("kubectl.views.services")
 
 local col_indices = { 1, 2 }
 api.nvim_buf_set_keymap(0, "n", "g?", "", {
@@ -23,9 +26,17 @@ api.nvim_buf_set_keymap(0, "n", "g?", "", {
         .. hl.symbols.clear
         .. "desc | "
         .. hl.symbols.pending
-        .. "<cr> "
+        .. "<1> "
         .. hl.symbols.clear
-        .. "containers",
+        .. "deployments | "
+        .. hl.symbols.pending
+        .. "<3> "
+        .. hl.symbols.clear
+        .. "configmaps | "
+        .. hl.symbols.pending
+        .. "<4> "
+        .. hl.symbols.clear
+        .. "secrets",
     })
   end,
 })
@@ -48,14 +59,6 @@ api.nvim_buf_set_keymap(0, "n", "d", "", {
     else
       api.nvim_err_writeln("Failed to describe pod name or namespace.")
     end
-  end,
-})
-
-api.nvim_buf_set_keymap(0, "n", "<bs>", "", {
-  noremap = true,
-  silent = true,
-  callback = function()
-    deployment_view.Deployments()
   end,
 })
 
@@ -92,6 +95,42 @@ api.nvim_buf_set_keymap(0, "n", "R", "", {
   silent = true,
   callback = function()
     pod_view.Pods()
+  end,
+})
+
+api.nvim_buf_set_keymap(0, "n", "1", "", {
+  noremap = true,
+  silent = true,
+  desc = "Deployments",
+  callback = function()
+    deployments_view.Deployments()
+  end,
+})
+
+api.nvim_buf_set_keymap(0, "n", "3", "", {
+  noremap = true,
+  silent = true,
+  desc = "Configmaps",
+  callback = function()
+    configmaps_view.Configmaps()
+  end,
+})
+
+api.nvim_buf_set_keymap(0, "n", "4", "", {
+  noremap = true,
+  silent = true,
+  desc = "Secrets",
+  callback = function()
+    secrets_view.Secrets()
+  end,
+})
+
+api.nvim_buf_set_keymap(0, "n", "5", "", {
+  noremap = true,
+  silent = true,
+  desc = "Services",
+  callback = function()
+    services_view.Services()
   end,
 })
 
--- a/ftplugin/k8s_pods.lua
+++ b/ftplugin/k8s_pods.lua
@@ -1,5 +1,6 @@
 local api = vim.api
 local container_view = require("kubectl.views.containers")
+local deployment_view = require("kubectl.views.deployments")
 local hl = require("kubectl.actions.highlight")
 local loop = require("kubectl.utils.loop")
 local pod_view = require("kubectl.views.pods")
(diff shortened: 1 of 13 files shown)


Ramilito:
I moved some of the global mappings into its own file under mappings.lua, how about registering those over there? So we don't need to repeat them on every view!
Also, I think the hints could be added globally as well, the code isn't optimized for it yet but you could just add it in utils/tables.lua somewhere under the other globals at row: 58, we can break it out into another file later if it's too ugly but at least we wouldn't need to repeat our selves

idelice:
> I moved some of the global mappings into its own file under mappings.lua, how about registering those over there? So we don't need to repeat them on every view! Also, I think the hints could be added globally as well, the code isn't optimized for it yet but you could just add it in utils/tables.lua somewhere under the other globals at row: 58, we can break it out into another file later if it's too ugly but at least we wouldn't need to repeat our selves

I'm getting following error after you merged main into my branch:

```lua 
  Error  21:47:57 msg_show.emsg E5108: Error executing lua: ...ce/projects/kubectl.nvim/lua/kubectl/views/pods/init.lua:1: loop or previous error loading module 'kubectl.resourcebuilder'
stack traceback:
	[C]: in function 'require'
	...ce/projects/kubectl.nvim/lua/kubectl/views/pods/init.lua:1: in main chunk
	[C]: in function 'require'
	/Users/delice/projects/kubectl.nvim/lua/kubectl/init.lua:8: in function 'open'
	/Users/delice/.config/nvim/lua/plugins/kubectl.lua:10: in function </Users/delice/.config/nvim/lua/plugins/kubectl.lua:9>
   Error  21:47:57 notify.error lazy.nvim Failed to run `config` for kubectl.nvim

.../delice/projects/kubectl.nvim/lua/kubectl/views/init.lua:1: loop or previous error loading module 'kubectl.resourcebuilder'

# stacktrace:
  - ~/projects/kubectl.nvim/lua/kubectl/views/init.lua:1
  - ~/projects/kubectl.nvim/lua/kubectl/utils/tables.lua:5
  - ~/projects/kubectl.nvim/lua/kubectl/resourcebuilder.lua:5
  - ~/projects/kubectl.nvim/lua/kubectl/views/deployments/init.lua:1
  - ~/projects/kubectl.nvim/lua/kubectl/completion.lua:1
  - ~/projects/kubectl.nvim/lua/kubectl/init.lua:15 _in_ **setup**
  - ~/.config/nvim/lua/plugins/kubectl.lua:16 _in_ **config**
```

lua/kubectl/actions/commands.lua
```diff
@@ -14,7 +14,7 @@ function M.shell_command_async(cmd, args, callback)
       table.insert(result, data)
     end,
     on_stderr = function(_, data)
-      callback(data)
+      vim.notify(data)
```
Ramilito, line 17: Great idea!

ftplugin/k8s_pods.lua
```diff
@@ -51,14 +62,6 @@ api.nvim_buf_set_keymap(0, "n", "d", "", {
   end,
 })
 
-api.nvim_buf_set_keymap(0, "n", "<bs>", "", {
```
Ramilito, line 54: This one is needed if you want to navigate up from pod -> deployment

Ramilito:
Do you have any 

> > I moved some of the global mappings into its own file under mappings.lua, how about registering those over there? So we don't need to repeat them on every view! Also, I think the hints could be added globally as well, the code isn't optimized for it yet but you could just add it in utils/tables.lua somewhere under the other globals at row: 58, we can break it out into another file later if it's too ugly but at least we wouldn't need to repeat our selves
> 
> I'm getting following error after you merged main into my branch:
> 
> ```lua
>   Error  21:47:57 msg_show.emsg E5108: Error executing lua: ...ce/projects/kubectl.nvim/lua/kubectl/views/pods/init.lua:1: loop or previous error loading module 'kubectl.resourcebuilder'
> stack traceback:
> 	[C]: in function 'require'
> 	...ce/projects/kubectl.nvim/lua/kubectl/views/pods/init.lua:1: in main chunk
> 	[C]: in function 'require'
> 	/Users/delice/projects/kubectl.nvim/lua/kubectl/init.lua:8: in function 'open'
> 	/Users/delice/.config/nvim/lua/plugins/kubectl.lua:10: in function </Users/delice/.config/nvim/lua/plugins/kubectl.lua:9>
>    Error  21:47:57 notify.error lazy.nvim Failed to run `config` for kubectl.nvim
> 
> .../delice/projects/kubectl.nvim/lua/kubectl/views/init.lua:1: loop or previous error loading module 'kubectl.resourcebuilder'
> 
> # stacktrace:
>   - ~/projects/kubectl.nvim/lua/kubectl/views/init.lua:1
>   - ~/projects/kubectl.nvim/lua/kubectl/utils/tables.lua:5
>   - ~/projects/kubectl.nvim/lua/kubectl/resourcebuilder.lua:5
>   - ~/projects/kubectl.nvim/lua/kubectl/views/deployments/init.lua:1
>   - ~/projects/kubectl.nvim/lua/kubectl/completion.lua:1
>   - ~/projects/kubectl.nvim/lua/kubectl/init.lua:15 _in_ **setup**
>   - ~/.config/nvim/lua/plugins/kubectl.lua:16 _in_ **config**
> ```

Weird!
Checked out your fork/branch and not getting that error, could you have some local changes that isn't the branch?

idelice:
> I moved some of the global mappings into its own file under mappings.lua, how about registering those over there? So we don't need to repeat them on every view! Also, I think the hints could be added globally as well, the code isn't optimized for it yet but you could just add it in utils/tables.lua somewhere under the other globals at row: 58, we can break it out into another file later if it's too ugly but at least we wouldn't need to repeat our selves

After moving those mappings to mappings.lua, it now only works once. If i press any of other shortcuts, nothing happens.. i really don't know what's happening lol

Ramilito:
Push those changes and I'll give it a try!

idelice:
> Push those changes and I'll give it a try!

done

Ramilito:
That was on me, just doing a require wasn't enough, since I'm trying not to pollute our mappings outside the plugin, we are setting them for each buffer (deployment, pods, configmaps each have their own buffer with the k8s_ prefix) but the require was only running on initial load and when switching buffer it stopped working.
Pushed a fix!

lua/kubectl/mappings.lua
```diff
@@ -41,5 +48,84 @@ function M.register()
...
+    noremap = true,
+    silent = true,
+    desc = "Services",
+    callback = function()
+      services_view.Services()
+    end,
+  })
+
+  vim.api.nvim_buf_set_keymap(0, "n", "g?", "", {
+    noremap = true,
```
Ramilito, line 98: We should probably do a merge of view specific + globals for the g?

idelice: I had specific hints g? for each view before migrating them to mappings.lua, but what exactly do you have in mind?

Ramilito: Similar to the normal hints section where we have view specific things and then we add global hints.
So for the pod specific view it would look like:
pod specific: [logs | desc | top | enter] - [1 pod | 2 dep | 3 configmaps]
service specific: [desc | port-forward] - [1 pod | 2 dep 3 confimaps]

Those globals would be added for every view but, I think it's really helpful if we could have contextual hints (similar to neogit)

idelice: I just checked the mapping for each view, and they're already showing their specific mappings under g?.

I think the next step is to add more functionality and later populate the g? for more hints if necessary. I think we can merge this PR since the implementation is in-scope.

Ramilito: Yeah, they get replaced by the global one but I agree that this PR has nothing to do with that, I'll create another PR that addresses that part, great work on the numbering shortcuts! I'm already using them 👍

Ramilito (approved):

PR #44, opened 2024-06-12
feat: map "r" to restart deployment

idelice:
Please give feedback on how to make it prettier if needed 😆


Ramilito:
The idea is solid, only feedback I got is that it would be nice to abstract away the menu behind a command instead of defining it, but! I don't feel that we need a plugin yet to handle our menus, I might reconsider that in the future if it becomes too difficult to manage.
I created another PR #49 that does the same thing you did but without the nui library and different keybinding for the confirmation, it's a bit more code for us to manage but not that much imo.

PR #54, opened 2024-06-15
feat: set image with "i" keybinding

idelice:
This only works *if* there is only *one* container and that image is the same as the deployment.

Please suggest feedback on how to make this more dynamic if needed


Ramilito:
I think this is a great first iteration! I'll setup an issue about making it handle multiple images and how that should look like.

Sorry for the late reply, I was trying to think about how we should handle the shortcuts and if we are going in correct direction. I'm conflicted about us overriding default navigation keys (for example the l to open logs), thinking of alternative ideas, such as opening a float with options instead, do you have any ideas? how about brainstorming here: #56

PR #55, opened 2024-06-16
[Feature] Use kubectl proxy

Ramilito:
Using kubectl proxy for subsequent curl calls to reduce latency.
With proxy:
- View Pods: avg 24ms - 60ms 
Without proxy:
- View Pods: avg 180 - 200ms


idelice:
I can't wait to get this one merged in :)

Ramilito:
> I can't wait to get this one merged in :)

Yeah me too, I do think the gain is worth the added complexity and curl dependency (most of us have curl anyways I believe), the gains seem to be even more when having a lot of pods, ~600 pods normally takes 300-350ms using kubectl but using curl is around 130-150ms.
I did try to do something about the json decoding but it seems to be too difficult to get any gains there. I've got some refactoring left to make it pretty but should be done soon!

I created a release (and tag) 0.1.0 in preparation so if this PR when merged isn't working out, people should be able to go back to a working version.

idelice:
I'm also positive that the performance issue I've experienced with this plugin was due to EKS (aws).

When using this plugin, with no EKS, it's fast. Otherwise, it gets pretty slow..

Ramilito:
Yeah I was kinda hoping to solve that using this approach, my theory is that having done the authentication (which I remember could take a while) once during startup, then reusing it for the requests after should shave off a lot of time 🤞

PR #94, opened 2024-07-07
remove gifs

mosheavni:
gifs are heavy on the repo
reference:
https://www.reddit.com/r/neovim/comments/wsg5yx/attention_to_all_plugin_authors_newcomers_careful/


Ramilito:
Nice catch, I didn't see that they were that big 😅

PR #95, opened 2024-07-07
feat: add kubectx command

Ramilito:
Adds ability to change context from within neovim with a kubectx command, maybe we should have a way to do it in the ui as well, not sure about that.


Ramilito:
kubectx

PR #119, opened 2024-07-21
feat: better keymaps

Ramilito:


Ramilito:
Heads up! @mosheavni @idelice redid almost all the mappings! (the hints are updated as well so should be apparent what the new ones are)

Issue #66, opened 2024-06-28
feat: port forward


Ramilito:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

- [x] be able to port-forward
- [x] be able to close port-forward
- [x] see what port-forwards are currently active

Ramilito:
#122 solves this

PR #122, opened 2024-07-22
feat: port forward mark and stop

Ramilito:
- [x] Add a symbol and hl to display where a pf exists
- [x]  When killing a pod, also close the pf
- [x] Add way to stop pf without killing the resource

You can now use gP to list all active port-forwards and close them without affecting the underlying service

Solves #66 

demo: 

pf

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,11 @@ Processes kubectl outputs to enable vim-like navigation in a buffer for your clu
   <summary>Diff view: <code>:Kubectl diff (path)</code></summary>
   <img src="https://github.com/user-attachments/assets/52662db4-698b-4059-a5a2-2c9ddfe8d146" width="700px">
 </details>
-   
+<details>
+  <summary>Port forward</summary>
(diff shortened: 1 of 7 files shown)


mosheavni:
just one note, the state of ports forwarded must be persistent across sessions, so in-memory is not good, need to save it to a file or dynamically check

Ramilito:
> just one note, the state of ports forwarded must be persistent across sessions, so in-memory is not good, need to save it to a file or dynamically check

Agreed! This implementation would be classified as a dynamic check since it's done on every redraw and does a lookup in the system (not memory) to check for any port-forwards, or did you mean something else?

mosheavni:
> > just one note, the state of ports forwarded must be persistent across sessions, so in-memory is not good, need to save it to a file or dynamically check
> 
> Agreed! This implementation would be classified as a dynamic check since it's done on every redraw and does a lookup in the system (not memory) to check for any port-forwards, or did you mean something else?

no, you got it :)

PR #127, opened 2024-07-25
feat: dynamic size on floating buffer

Ramilito:


mosheavni:
LGTM! tested locally and the floats look way better.

Ramilito:
It has a bug that I couldn't solve without resolving the preloading buffer hack where if you change the size of the content there would be a background buffer that started showing. Will try to solve both this and the hack in this pr #145

Ramilito:
Closing this one since I did it in #145

Issue #135, opened 2024-08-04
Implement generic resource view


jemag:
**Is your feature request related to a problem? Please describe.**
I would like to be able to view any resource on the cluster.

**Describe the solution you'd like**
Have a generalized view where you can query (if possible with autocomplete), any resource that is available through `kubectl api-resource`. 

This would allow seeing more niche standard resources as well as any custom resource (e.g., external-secrets, servicemonitors, policyreports, argocd applications, etc.)

**Additional context**
I realise that the following issue https://github.com/Ramilito/kubectl.nvim/issues/23 already exists and mentions CRDs, but I think this feature is more general and would not only be for custom resource definitions, but any resource/custom resource whatsoever.

Ramilito:
Hi and thanks for creating an issue 🙏!

Would love to hear more on how this should look like and would it be different from the Fallback view mentioned in #23?
The idea with that one is to handle the output from example :Kubectl get external-secrets and give some sort of basic functionality.

jemag:
I am not quite sure how the fallback view would work/look like, but, personally, I would rather avoid typing the whole `kubetctl get XXXX` command to trigger this.

I guess what I am thinking is to have a hotkey trigger something similar to the `<c-f>` Filter, where you could enter the resource name with auto-complete (auto-complete source could be retrieved from `kubectl api-resources`). Once you enter the resource you desire, you get a view similar to the pods/deployments/secrets one, where it just list all the resources and you could `describe`, `delete`, `edit` any of them.

Basically, a bit similar to the k9s:
![image](https://github.com/user-attachments/assets/edfc5efe-a8a5-4759-a520-4b56a7e65f4d)
and
![image](https://github.com/user-attachments/assets/3577411f-6b03-480b-a6a4-7cf67457fb06)


Hopefully that made sense, let me know if it doesn't.

Ramilito:
I loves the suggestion and really happy with the results, I just merged it and would love to hear if it meets your expectations!

You trigger the search using ctrl+a (short for alias), start typing and tab to cycle through the suggestions, if it's a view we support then it should load those otherwise fallback on the generic view.

The generic view is pretty barebones atm, it has name and namespace always and some other columns if some conditions are met. 
My thoughts regarding that is to let it grow organically and try to create real views when we need it instead of covering all scenarios.

mosheavni:
> I loved the suggestion and really happy with the results, I just merged it and would love to hear if it meets your expectations!
> 
[...]

PR #136, opened 2024-08-05
feat: aliases view

Ramilito:
Solves #135 and parts of #23

Adds an alias search pop with tab autocomplete.
Choosing one will load that view.
If we have support for the view, that will be loaded otherwise a fallback view (generic view) with basic functionality (describe, edit, delete)

aliases


mosheavni:
Does every <kbd><C-a></kbd> reloads the api-resources?
I'm not sure the flag `--cached` is working for me, hitting <kbd><C-a></kbd> is kinda slow every time.

Ramilito:
Hmm I was hoping the cached would solve the performance tbh, running my own test seems to be performant, could you try running these commands?
```
time kubectl api-resources --cached
```
```
time kubectl api-resources
```

For me, the difference is almost 150ms so it seems to be doing something, the timing inside the plugin is also similar in loading time, from:
Execution time: 0.086942375 seconds  <- with cached
Execution time: 0.176638708 seconds   <- without cached

mosheavni:
> Hmm I was hoping the cached would solve the performance tbh, running my own test seems to be performant, could you try running these commands?
> 
> ```
> time kubectl api-resources --cached
> ```
> 
> ```
> time kubectl api-resources
> ```
> 
> For me, the difference is almost 150ms so it seems to be doing something, the timing inside the plugin is also similar in loading time, from: Execution time: 0.086942375 seconds <- with cached Execution time: 0.176638708 seconds <- without cached

i have a cluster with many CRDs, a heavy cluster.
it gives me:

```
kubectl api-resources --cached  0.24s user 0.07s system 15% cpu 1.996 total
```

for me it's noticable for a window that should open very quick (maybe also rendering the float takes time)

I think it's better to keep an in-memory cache (`M.cached_api_resources = {...}`)

Ramilito:
Added a new issue: #139

Issue #144, opened 2024-08-08
Base64 decode secrets


RobinCamarasa:
First of all thank you for the amazing plugin!

**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I describe a secret (using `gd`) and can't access the content.
![image](https://github.com/user-attachments/assets/d5c5ac05-eb72-499f-8e50-a7f4e85ebf23)

**Describe the solution you'd like**
I would like that there is a mapping to display the secret and `base64 --decode` it.

**Describe alternatives you've considered**
This adds a dependency on base64 but already having the base64 encoded secret would also help a lot.

**Additional context**
/

Ramilito:
Thank you for the suggestion!
Found a lua decoder code so we managed to avoid the base64 dependency, and it seems to be working great on all the secrets I have!

What I did was to change the describe to a get -o yaml to get the encoded printed in our describe so it's visible now and not just a count of bytes.
Then I added a keymap where you only need to get to the line of the encoded secret and press enter to decode it, you can be anywhere on that line.
The decoded value is then printed to vim.notify.

Wasn't 100% on how to solve it UX wise so if you have any feedback I'd love that.

RobinCamarasa:
Thanks it is already really great (except copying the secret to the clipboard found a workaround with :message but it is not ideal). My suggestion for the UX would be to toggle inplace the base 64 encode and decode with `<cr>` key binding, this would make it easier to copy the secret to the clipboard and manipulate it using nvim. I have however no idea of how easy it is to implement.

Ramilito:
Yeah that would be really great, I'l give it a shot but tbh it sounds a bit challenging

Ramilito:
The toggle feature was a bit too much, figuring out if a text is decoded or encoded or keeping track of them isn't worth the hassle imo.
Settled on a compromise to only decode and replace the current encoded value, if you want to revert it just close and open the description window.
If that's not enough, we might revisit it later, but I feel that it's pretty good where it's at right now!

PR #146, opened 2024-08-08
feat: decode base64 secrets

Ramilito:
Added a base64 utility decoder.
Added ft type for secret_desc
Added enter keymap to decode line

Solves #144


mosheavni:
why not use `vim.base64`?
https://neovim.io/doc/user/lua.html#_vim.base64

Ramilito:
> why not use `vim.base64`? https://neovim.io/doc/user/lua.html#_vim.base64

lol indeed why not!
Removed that horrible base64 decoder and using vim.base64 instead, thank you for the correction 🙏

PR #159, opened 2024-08-12
fix: manipulate copy instead of original

Ramilito:
#158 Introduced a bug since lua doesn't send tables by value and instead uses reference, that was fine when we didn't have the values stored in a variable (hardcoded parameters) but started getting corrupt when sent into a function that manipulates the input parameter.

Solved it by making copies of the arguments but the correct solution is to use immutable tables for readonly values.


mosheavni:
sorry!

PR #169, opened 2024-08-15
fix/update-context-on-kubectx

### LINKED ISSUE #170: Changing the context doesn't change the header
When changing the context using `:Kubectx` it doesn't change the context header with the context, namespace, user and cluster
<kbd>gr</kbd> doesn't trigger the `context.setup()`

mosheavni:
Fixes #170 
When changing the context, the context info at the top stays the same because it is only configured by `state.setup()`
so who not call it whenever changing context?


Ramilito:
Nice catch! :shipit:

Ramilito (approved):

PR #171, opened 2024-08-16
feat: use informer to update state

Ramilito:
Investigate if the informer can be something to increase performance and reduce api calls.
Motivation: This should make very large cluster fesable since we would only be getting large amount of data once and then listen to changes.

This is now working as expected, there was an issue with libuv buffer size causing stdout to split into multiple lines but think that is sorted with a leftover mechanic and a queue mechanic
The performance gains are huge but also instead of refreshing on a set interval, which makes us miss certain events, we now get all intermediate states and statuses.

TODO:
- [x] Not handling bulk changes
- [x] Handle incomplete json caused by libuv
- [x] Create a queue to handle events in our own pace
- [x] Make it work for other views in a generic way
- [x] A lot of jumping around when changes happen (sorting seems to help with this, so adding a default sort, not sure if it's enough though)
- [x] PF not great
- [x] Long session disconnects curl
- [x] Stability check/verification
- [x] Sometimes the informer doesn't shutdown when moving to another view

BONUS:
- [x] Instant sort, now that we have a redraw without fetching everything, we can sort really fast
- [x] Filter doesn't run unnecessarily


mosheavni:
a lot of code and honestly I did not really understand the description.
Can you explain to me like a 10 year old what this feature is?

Ramilito:
I'm so happy that you ask!
The a lot of code is because I made the mistake of thinking I would be done sooner with this, so when I found an issue/bug instead of creating a new branch and merge that I just included it in this one (won't do that again...).
So a bunch of changes aren't really super relevant to this task.

The most important parts are the split of initializing and redrawing a view.
If you look at the pods init.lua file you can see the old ```M.View``` function and a new ```M.Draw```


The view function does a fetch like before and then it starts a watcher (informer pattern), when the watcher is running we will be notified on changes to the resources in kubernetes, there are three types of changes ADDED, MODIFIED, DELETED.

And since we have the state in memory we only need to update a specific item which will result in us only needing to redraw the items to the buffer again.

The draw is super fast and I have my refresh rate down to 50ms and everything is super responsive, the sorting for example used to do a fetch but now it only needs to sort items in memory.

mosheavni:
> I'm so happy that you ask! The a lot of code is because I made the mistake of thinking I would be done sooner with this, so when I found an issue/bug instead of creating a new branch and merge that I just included it in this one (won't do that again...). So a bunch of changes aren't really super relevant to this task.
> 
> The most important parts are the split of initializing and redrawing a view. If you look at the pods init.lua file you can see the old `M.View` function and a new `M.Draw`
> 
> The view function does a fetch like before and then it starts a watcher (informer pattern), when the watcher is running we will be notified on changes to the resources in kubernetes, there are three types of changes ADDED, MODIFIED, DELETED.
> 
> And since we have the state in memory we only need to update a specific item which will result in us only needing to redraw the items to the buffer again.
> 
> The draw is super fast and I have my refresh rate down to 50ms and everything is super responsive, the sorting for example used to do a fetch but now it only needs to sort items in memory.

Thanks for the explanation. definitely makes sense.

PR #172, opened 2024-08-16
feat: batched fetch

Ramilito:
Investigate if splitting the result into batches is better for user experience

The poc is working but there are some issues to be resolved

- [ ] Column width is calculated based on the content of the rows, when batching, that calculation is not correct (might be solved with #171 where redraw is cheap)
- [ ]  Only really worth it on huge datasets 1k+ rows (especially with #171 )


Ramilito:
I think this is a bad approach, we should probably just set a limit to something we believe we can handle and have good filtering if the results exceed that.

PR #180, opened 2024-08-18
Feat/custom resources

### LINKED ISSUE #178: feat: fallback with custom fields
There are alot of resources I get on a daily basis and want to see some more fields.
Instead of creating a dedicated view for each one, it would be amazing if I can add some custom fields in the setup and it would be read when rendering a fallback view

something like:
```lua
require('kubectl').setup {
  custom_views = {
    resource = "externalsecret",
    display_name = "ExternalSecrets",
    ft = "k8s_externalsecret",
    url = { "{{BASE}}/apis/external-secrets.io/v1beta1/externalsecrets" },
    headers = {
      NAMESPACE = '', -- part of fallback view, no need to recalculate
      NAME = '', -- part of fallback view, no need to recalculate
      STORE = '.spec.secretStoreRef.name',
      ['REFRESH INTERVAL'] = '.spec.refreshInterval',
      STATUS = '.status.conditions[0].reason',
      READY = '.status.conditions[0].type',
      AGE = '' -- part of fallback view, no need to recalculate
    }
  }
}
```

I think we might move `viewsTable` to the definition of each view as `aliases = { ... }`

I'll try to contribute here, just let me know anything I need to know to do that.

mosheavni:
fixes #178 

While this is working, I am sure you have to say about the code something @Ramilito 
open to suggestions.


mosheavni:
I see that the headers are persisting when viewing a different fallback

Ramilito:
While this is a great idea I'm not  sure about the approach.

I think I need to think a bit on it before merging this

mosheavni:
I see now that you can extract what are the printed fields for a CRD. Maybe this direction?

mosheavni:
Example
if we add this header to the curl request:
`Accept: application/json;as=Table;v=v1;g=meta.k8s.io,application/json;as=Table;v=v1beta1;g=meta.k8s.io,application/json`
Then we get 2 things:
`.columnDefinitions` and `.rows[*].cells`

We can use that to build the fallback resource

```
curl \
  -H 'Accept: application/json;as=Table;v=v1;g=meta.k8s.io,application/json;as=Table;v=v1beta1;g=meta.k8s.io,application/json' \
  'http://localhost:8001/apis/external-secrets.io/v1beta1/externalsecrets?limit=500'
```

image

The problem I see is that the fallback uses `kubectl` and not `curl` since we don't know if the user requested a short name of the resource name in a format that the request via curl would accept.
Thoughts? @Ramilito

Ramilito:
Yeah exactly! I was thinking of two possible ways, one is to let the user use ftplugin in some way to register a view instead of the fallback view. 
The other is to use the crd, I played with it when creating the fallback views but scoped it down to what we have right now.

We should be able to move to curl for the fallback as well.
The short names, not sure how that is to be solved without user configuration.

mosheavni:
Let's say we ignore the short names and provide that functionality only to resource name as seen in 'kubectl api-resources'
How can we also get the api version for that api-resource? We must have it for the full URI params?

Ramilito:
You should be able to find something when curling the /apis endpoint and then looking for preferred version (since we can have multiple crd versions) and the name of the group.

mosheavni:
> You should be able to find something when curling the /apis endpoint and then looking for preferred version (since we can have multiple crd versions) and the name of the group.

should we do it here:
https://github.com/Ramilito/kubectl.nvim/blob/62db8f047a8a859f8f9b79427e76d26aa8f81ca6/lua/kubectl/views/init.lua#L15-L20

or on the fallback view?

Ramilito:
I think there is great!
If the cached_api_resources.value could be a table with ```{ name = foo, url = "api_endpoint/version"}``` that would work really well I think

mosheavni:
> I think there is great! If the cached_api_resources.value could be a table with `{ name = foo, url = "api_endpoint/version"}` that would work really well I think

I'll try to implement something.

Issue #174, opened 2024-08-18
add `close` and `toggle` methods


mosheavni:
right now we open the plugin with `<cmd>lua require("kubectl").open()<cr>`
and then we map <kbd>\<leader>k</kbd> to close the plugin.
it would be better to not define:
https://github.com/Ramilito/kubectl.nvim/blob/f6575b9b7f96640195953c89b2d87f594220a124/lua/kubectl/mappings.lua#L10
and just allow the user to map his own `<cmd>lua require("kubectl").toggle()<cr>`
and add a `require("kubectl").close()`, again to allow the user for customization.

Ramilito:
Yeah that might be more intuitive.
Just to be clear, are you suggesting we expose three functions ```open ``` ```close``` ```toggle``` or just ```open``` and ```close```?

mosheavni:
> Yeah that might be more intuitive. Just to be clear, are you suggesting we expose three functions `open ` `close` `toggle` or just `open` and `close`?

yes all three, that's the way most plugins work and I think it is more intuitive as you stated.

halshar:
yes, three would be better. Right now I've to open the the tool in another vim buffer as its easier for me to switch buffers than close and open the tool, toggle option would be much helpful

Ramilito:
This is fixed in #183, thank you all for the awesome suggestions ❤️

PR #183, opened 2024-08-19
feat: different close and open strategy

Ramilito:
Solves #174

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,17 @@ return {
 
 ## ⌨️ Keymaps
 
+We expose open, close and toggle to bind against, example alternatives below:
+
+#### Toggle
 ```lua
--- Recommended is to have the same open mapping as your close (```<leader>k```) the plugin for a toggle effect.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -79,9 +79,17 @@ return {
...
+#### Toggle
+```lua
+vim.keymap.set("n", "<leader>k", '<cmd>lua require("kubectl").toggle()<cr>', { noremap = true, silent = true })
+```
+
+#### Open with ```<leader>k``` and close with ```q```
 ```lua
--- Recommended is to have the same open mapping as your close (```<leader>k```) the plugin for a toggle effect.
 vim.keymap.set("n", "<leader>k", '<cmd>lua require("kubectl").open()<cr>', { noremap = true, silent = true })
+vim.keymap.set("n", "q", '<cmd>lua require("kubectl").close()<cr>', { noremap = true, silent = true })
```
mosheavni, line 92: i think it's better to give an example only for the `toggle()` case since it's most common.
Users can choose when and if they ever need the `open()` and `close()` but nonetheless have the option to use it.

Ramilito: That's fair, changed it! Thanks for the review 🙏

PR #187, opened 2024-08-20
Feat/custom resources

### LINKED ISSUE #178: feat: fallback with custom fields

mosheavni:
fixes #178 
Trying this again, I do need help.
The fallback view is working great

- [x] Completion is not working yet, for some reason I cannot use `vim.print()` to debug, is there a better way?
- [x] When getting the API resources, get both singular and plural names (i.e `deployment` and `deployments`)
- [x] Cannot sort on fallback view
- [ ] Replace the `viewsTable` with the cached shortNames

in the current situation, fallback view is much more informative:
image

But until the bullets I specified are sorted-out, we better not merge it.

Can you help me with the following @Ramilito ?


Ramilito:
That looks great! I will take a look at the code and help out as soon as I can but super impressive results!

mosheavni:
> That looks great! I will take a look at the code and help out as soon as I can but super impressive results!

if you get a chance, the most critical things for this PR are:
- Sorting by headers
- <kbd>\<C-a></kbd> aliases menu completion is not working  - I'm having trouble debugging this part, I cannot print anything maybe since we're in insert mode? or because we're on a vim.schedule block?

Ramilito:
> > That looks great! I will take a look at the code and help out as soon as I can but super impressive results!
> 
> if you get a chance, the most critical things for this PR are:
> 
> * Sorting by headers
> * <C-a> aliases menu completion is not working  - I'm having trouble debugging this part, I cannot print anything maybe since we're in insert mode? or because we're on a vim.schedule block?

Hmm...how are you printing? Is it that vim.print  and notification in the completion function? Are you pressing tab when testing it? tbh it looks correct 😕

.luarc.json
```diff
@@ -1,4 +1,4 @@
 {
-    "Lua.runtime.version": "Lua 5.1",
-    "Lua.diagnostics.globals": ["vim"],
+  "Lua.runtime.version": "Lua 5.4",
```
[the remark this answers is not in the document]
mosheavni: yeah bummer. didn't find a good alternative for goto other than messy if else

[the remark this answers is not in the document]
mosheavni: fixed. @Ramilito  can you check?

lua/kubectl/actions/informer.lua
```diff
@@ -60,6 +60,9 @@ local function decode_json_objects(json_strings)
 end
 
 local function process_event(builder, event)
+  if not event.type then
```
Ramilito, line 63: Did you have an error with event.type here? It shouldn't be happening since all the events that get here are from the kubernetes watch event, could you print out the event?

mosheavni:
fixed sorting, fixed luajit version, everything is working.

Ramilito:
> fixed sorting, fixed luajit version, everything is working.

Just tested it out and agree! it seems to be working really well, do you mind if I do some refactoring before merging? Just simple stuff

lua/kubectl/actions/informer.lua
mosheavni: the problem is where you append `?watch=true&resourceVersion=...`
On custom resources, this produces `"404 page not found"`
I found this out after merging the informer changes to my branch.

I don't think `watch` and `resourceVersion` is applicable for any resource, so this is to be able to fallback

mosheavni:
> > fixed sorting, fixed luajit version, everything is working.
> 
> Just tested it out and agree! it seems to be working really well, do you mind if I do some refactoring before merging? Just simple stuff

yes go ahead!

mosheavni:
one small issue with sorting, when the field name has spaces (like `REFRESH INTERVAL`) the sorting does not work.
any suggestions for this?

mosheavni:
namespace filter is also not working.. :(

lua/kubectl/actions/informer.lua
Ramilito: hmm that could be true, do you have a crd I could test it out with?

mosheavni:
> namespace filter is also not working.. :(

fixed in https://github.com/Ramilito/kubectl.nvim/pull/187/commits/8ad391d959acc7c8d221e4ece652d015266a6c9f

lua/kubectl/actions/informer.lua
Ramilito: nvm got it!

lua/kubectl/views/fallback/init.lua
```diff
@@ -23,19 +23,39 @@ end
...
 
+  -- default fallback values
   definition.resource = M.resource
   definition.display_name = M.resource
   definition.url = get_args()
   definition.ft = "k8s_fallback"
   definition.hints = {
     { key = "<gd>", desc = "describe", long_desc = "Describe selected " .. M.resource },
   }
+  definition.cmd = "kubectl"
```
Ramilito, line 40: Do you think we should still support the kubectl variant? Wont the curl work for all use cases?

lua/kubectl/actions/informer.lua
Ramilito: So what happened was that the crds url was different and that's why we didn't get the events, after that the issue became the difference in tables between crds and normal view. That's been fixed now!

.luarc.json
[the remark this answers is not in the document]
Ramilito: Looking great!

Ramilito:
> one small issue with sorting, when the field name has spaces (like `REFRESH INTERVAL`) the sorting does not work. any suggestions for this?

Fixed this one, it's almost perfect, you have to be on the first word when sorting, so in the example above, standing on INTERVAL won't work but standing on REFRESH would work.

Ramilito:
I believe this one is ready to be merged, there is just the question if we want to support kubectl json in fallback view anymore or if we could clear it, what do you think? will this solution not work for all  custom views?

Ramilito (approved):

mosheavni:
good morning. wow how impressive!
learned alot from your commits.
unfortunately, sorting is not that great
when I sort by a column then by another column it actually messes up the resource, the namespace is wrong!

I have 1 ExternalSecret with `status: SecretSyncedError` the resource line is:
```
→ kg externalsecrets.external-secrets.io -A G -i error
ml                             sp-rmq-credentials                                param-store   5m                 SecretSyncedError   False
```

when first opening, then sorting by refresh interval, then refresh by status, the namespace has changed
```
rabbitmq                       sp-rmq-credentials                                param-store   5m                 SecretSyncedError   False   
```

the way to fix it is to sort by namespace then refresh with <kbd>gr</kbd>

Unfortunately I cannot record a demo gif since it might be sensitive

Ramilito:
hmm that's weird, I'll see if I can replicate it

lua/kubectl/views/fallback/init.lua
mosheavni: in a weird case when someone opens `kubectl` on nvim, then install some CRD outside of nvim, then this would allow the plugin to see the most basic fallback since cached_resources doesn't contain this new one.
very weirdly specific but not >=1% chances of happening

mosheavni:
> hmm that's weird, I'll see if I can replicate it

we can merge this one and open a separate ticket no? it's minor compared to the huge benefit this provides.

Ramilito:
> > hmm that's weird, I'll see if I can replicate it
> 
> we can merge this one and open a separate ticket no? it's minor compared to the huge benefit this provides.

That's a fair point, did however not read it before fixing the sort 😅 , will merge it now but sorting should be working for you now.
Changed the way we do it to be a lot more robust.

lua/kubectl/views/fallback/init.lua
Ramilito: Good point, lets leave it in for now until it becomes a maintenance issue!

Ramilito:
Really nice contribution btw! Do you mind me inviting you as a contributor?

PR #190, opened 2024-08-21
Feat/logging

mosheavni:
A nice logging factory thanks to none-ls

work in progress, not working currently

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,9 @@ Processes kubectl outputs to enable vim-like navigation in a buffer for your clu
 - kubectl
 - curl
 - neovim >= 0.10
- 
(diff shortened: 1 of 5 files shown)


Ramilito:
I'm not sure I understand what this is supposed to do but the strategy I'm opting for is to use vim.notify and then let the users them self choose a message handler! Unless this is super awesome so looking forward to seeing a screenshot :D.
On another note, I don't want to add plenary as a dependency, I don't really like it, lots of issues and not super well maintained.
Lots of the things it offered are now easy to do just using neovim +10, such as jobs and async stuff. I predict people will move away from it.

Issue #194, opened 2024-08-22
feat: add content header styling


Ramilito:
**Describe the solution you'd like**

The headers are only a different color today, explore if we can make them prettier and or more functional.

Ramilito:
Closing this since I can't figure out anything better than the one we have currently

PR #196, opened 2024-08-22
feat: style headers

Ramilito:
Add more styling to headers, was originally thinking about adding a divider but that didn't look good
Solves #194


Ramilito:
Underline makes it look like this:
image

PR #197, opened 2024-08-22
feat: use the full width of the column to sort on

Ramilito:
This allows for sorting in the full width of the  column, unfortunately it interferes with #196


Ramilito:
@mosheavni Check this one out! I believe you'll like it!
It allows us to sort from any row instead of going back to the header row!

mosheavni:
> @mosheavni Check this one out! I believe you'll like it! It allows us to sort from any row instead of going back to the header row!

Love it.

PR #198, opened 2024-08-22
feat/completion-and-namespace

mosheavni:
1. there was no completion for resources under `/api/v1` since it's not returned in `/apis` so fetch it before
2. remove `{{NAMESPACE}}` for cluster-scoped resources
3. Add missing `{{NAMESPACE}}` to service accounts and pvc

regarding 2:
image


Ramilito:
Nice!

PR #199, opened 2024-08-22
fix/filter

mosheavni:
filter was not working well

### DIFF
--- a/lua/kubectl/resourcebuilder.lua
+++ b/lua/kubectl/resourcebuilder.lua
@@ -157,9 +157,7 @@ function ResourceBuilder:process(processFunc, no_filter)
     return self
   end
 
-  local filter = state.getFilter()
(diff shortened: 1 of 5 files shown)


Ramilito:
Hmm if you don't need an exact match you should be able to send that in as a parameter? 
What's the filter parameter we are trying to  use

mosheavni:
In this case we better use 'string.find' instead of 'string.match'
Cause lua patterns come straight from hell
Previously, filter wouldn't work if you have dashes in your filter term.
string.find(t, string, 1, true) provides a literal search.

Ramilito:
That sounds reasonable and I love the simpler way!
Just make the test run properly, there is a failure here:
image

mosheavni:
> That sounds reasonable and I love the simpler way! Just make the test run properly, there is a failure here: image

fixed, thanks.

Issue #120, opened 2024-07-21
feat: override keymaps


Ramilito:
**Describe the solution you'd like**

Not sure here, the best would be to use the Plug feature.

**Describe alternatives you've considered**

The normal mappings setup

mosheavni:
`<Plug>` keymaps are the best way to go.

Ramilito:
This is solved in #273, we still need to update docs on how to use it.

Ramilito:
We have an issue to update docs so will close this one!

PR #203, opened 2024-08-22
feat/allow-overriding-keys

mosheavni:
provide an option to not set 0-5 keymaps


Ramilito:
I don't really like to add that kind of config, lets rethink how we want the mappings to be customized in #120

PR #205, opened 2024-08-22
fix/age-sort

mosheavni:
properly sort by age, parse timestamp, convert to epoch and use for comparison

that's it for tonight.. 😴

### DIFF
--- a/lua/kubectl/resourcebuilder.lua
+++ b/lua/kubectl/resourcebuilder.lua
@@ -197,7 +197,7 @@ function ResourceBuilder:sort()
         local isSortByA = type(valueA) == "table" and valueA.sort_by
         local isSortByB = type(valueB) == "table" and valueB.sort_by
         if isSortByA and isSortByB then
-          return comp(epochA, epochB)
+          return comp(valueA.sort_by, valueB.sort_by)
         elseif tonumber(valueA) and tonumber(valueB) then
           return comp(valueA, valueB)
         else
--- a/lua/kubectl/resourcebuilder.lua
+++ b/lua/kubectl/resourcebuilder.lua
@@ -210,6 +210,12 @@ function ResourceBuilder:sort()
         elseif tonumber(valueA) and tonumber(valueB) then
           return comp(valueA, valueB)
         else
+          -- sort by restarts
+          local numA = tonumber(tostring(valueA):match("(%d+)%s+%(.+ago%)"))
+          local numB = tonumber(tostring(valueB):match("(%d+)%s+%(.+ago%)"))
+          if numA and numB then
+            return comp(numA, numB)
+          end
           return comp(tostring(valueA), tostring(valueB))
         end
       end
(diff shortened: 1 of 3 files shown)


mosheavni:
also sort by restarts (of pod for example) in https://github.com/Ramilito/kubectl.nvim/pull/205/commits/434a54d75b67fa951914b024bb65aad5fbbd5d14

lua/kubectl/resourcebuilder.lua
```diff
@@ -196,13 +196,24 @@ function ResourceBuilder:sort()
         end
         if type(valueA) == "table" and type(valueB) == "table" then
           if valueA.timestamp and valueB.timestamp then
-            return comp(tostring(valueA.timestamp), tostring(valueB.timestamp))
```
Ramilito, line 199: I think we can do something prettier here, if you look at the utils.time.lua file, we do something like this with a timestamp:
```
  local parsed_time = vim.fn.strptime("%Y-%m-%dT%H:%M:%SZ", timestamp)

```

lua/kubectl/resourcebuilder.lua
```diff
@@ -196,13 +196,24 @@ function ResourceBuilder:sort()
...
+            local epochB = os.time({ day = dayB, month = monthB, year = yearB, hour = hourB, min = minB, sec = secB })
+            return comp(epochA, epochB)
           else
             return comp(tostring(valueA.value), tostring(valueB.value))
           end
         elseif tonumber(valueA) and tonumber(valueB) then
           return comp(valueA, valueB)
         else
+          -- sort by restarts
+          local numA = tonumber(tostring(valueA):match("(%d+)%s+%(.+ago%)"))
```
Ramilito, line 212: Hmm we already have a elseif that is supposed to handle numbers, instead of bloating this sort function, I would rather handle this in in a similar way as we do the timestamp.
In the timestamp we add another field to the value ```{ value= "some value", symbol = "some hl group", timestamp = "we sort on this" }```
We could do something similar when building the content rows, when creating the Age column value we would add something like this:
```{ value= "some value", symbol = "some hl group", sort_on = "A number" }```
You would then check this in an elseif after the timestamp check.

Another option, maybe for a future rewrite, is to have a function that this method invokes instead, i.e bring you own sort

Ramilito:
> that's it for tonight.. 😴

Was not expecting all these pr's at 4AM 🤣

lua/kubectl/resourcebuilder.lua
mosheavni: done

lua/kubectl/resourcebuilder.lua
mosheavni: Ok I committed https://github.com/Ramilito/kubectl.nvim/pull/205/commits/3725db3d5b625c647e35ad7b039e5130b2efd3be
can you please take a look @Ramilito  ?

Ramilito: Perfect!

Ramilito:
The type checker is complaining otherwise I love it!

Ramilito:
Looks amazing!

PR #211, opened 2024-08-23
fix: revert the exact filter method

Ramilito:
In #199 we removed the exact matching for filtering, this caused problems when searching for aliases, for example trying to go to rolebinding would match for clusterrolebinding.
Returned the exact matching so that doesn't happen.

@mosheavni Will this revert affect the issue you had in #199? I'm thinking the issue you had was for filtering in content and not when filtering aliases, correct?


mosheavni:
Since you kept `string.find` and not reverted back to `string.match` the filter still works! so all good.

PR #216, opened 2024-08-23
chore/get-shorter-completion-first

mosheavni:
When completing, get the shortest matching completion
for example:
`ingr` -> `ingresses`
instead of `ingr` -> `ingressclassparams`


Ramilito:
Just move it to utils/completion.lua and we can merge it 👍

mosheavni:
done.

Issue #214, opened 2024-08-23
feat: allow editing the yaml file in-place


mosheavni:
a super cool feature of k8s is that you can click <kbd>y</kbd> and you go to $EDITOR to edit the yaml.
Since we're already on vim, opening a terminal causes us to opening a vim inside vim (like inception).
We need to copy the logic behind `kubectl edit {resource}` to get the temporary file we're editing and maybe an autocommand for when the file is closed and saved, send that file to the server (I believe that's what they do behind the scenes).
WDYT @Ramilito ?

Ramilito:
Maybe ```ge``` isn't working for you? I'm using it quite often and haven't had issues with it:

![ge](https://github.com/user-attachments/assets/5a159c10-0ec2-4956-9485-f5fd4ced5933)

mosheavni:
> Maybe `ge` isn't working for you? I'm using it quite often and haven't had issues with it:
> 
[...]

PR #218, opened 2024-08-23
Feat/native kubectl edit

mosheavni:
fixes: #214 
Kubectl edit using nvim editor, no inception.
I think it's nice what I got so far:

Currently pending:

- [x] when closing the yaml file, it closes vim instead of going back to the previous view
- [x] implement `kubectl patch -f` to apply changes

@Ramilito would love your input here

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Processes kubectl outputs to enable vim-like navigation in a buffer for your clu
 </details>
 <details>
   <summary>Sort by headers</summary>
-    <sub>By moving the cursor to a header word and pressing <code>s</code></sub>
+    <sub>By moving the cursor to a column and pressing <code>s</code></sub>
   <img src="https://github.com/Ramilito/kubectl.nvim/assets/17252601/9f96e943-eda4-458e-a4ba-cf23e0417963" width="700px">
 </details>
 <details>
--- a/ftplugin/k8s_container_logs.lua
+++ b/ftplugin/k8s_container_logs.lua
@@ -11,6 +11,15 @@ local function set_keymaps(bufnr)
       container_view.tailLogs(pod_view.selection.pod, pod_view.selection.ns)
     end,
   })
+
+  vim.api.nvim_buf_set_keymap(bufnr, "n", "gw", "", {
+    noremap = true,
+    silent = true,
+    desc = "Toggle wrap",
+    callback = function()
+      vim.api.nvim_set_option_value("wrap", not vim.api.nvim_get_option_value("wrap", {}), {})
+    end,
+  })
 end
 
 --- Initialize the module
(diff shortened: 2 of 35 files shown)


Ramilito:
>  when closing the yaml file, it closes vim instead of going back to the previous view

I think you either add a key mapping that overrides `q` that only closes the buffer instead of the window. Or, open a floating buffer before you load the content, that would be similar to the current way we do it for describing a resource.

It's looking really good in the gif!

mosheavni:
> > when closing the yaml file, it closes vim instead of going back to the previous view
> 
> I think you either add a key mapping that overrides `q` that only closes the buffer instead of the window. Or, open a floating buffer before you load the content, that would be similar to the current way we do it for describing a resource.
> 
> It's looking really good in the gif!

creating a `cmap` for `:q` and `:wq` makes the command line hang and act weird.
I'm looking for a better way

mosheavni:
I decided to use a vsplit instead of changing the current buffer. it's working.
@Ramilito  maybe you want to move stuff around before we merge?

Ramilito:
I'd rather have a floating buffer for this, do you see any issues with that?

mosheavni:
> I'd rather have a floating buffer for this, do you see any issues with that?

I'm not against trying. wanna give the code a go?

Ramilito:
> > I'd rather have a floating buffer for this, do you see any issues with that?
> 
> 
> 
> I'm not against trying. wanna give the code a go?

I'll give it a try later tonight!

Ramilito:
Did some small changes:
- use the normal floating buffer as we use for the other operations.
- remove the temp file handling and just use in-memory, I believe it makes it easier and opens up another way for us to actually do an apply instead of being forced to save the buffer.

How about replacing the autocommands with a simple keybinding to apply the change? Not sure what's best there, we could also just mimic a "file save" if we think that's the best one

mosheavni:
image

few issues:
- [ ] no line numbers, no relative line numbers
- [ ] filetype should be "yaml" so yaml-ls would start
- [ ] yaml-ls works on actual files, so when the buffer is not saved, there's no completion.

Really editing a kubernetes yaml without schema completion is pointless for me.

mosheavni:
> Ah, mb, is that a hard requirement from lsp to have a real file? If so, we should revert the in memory one (I can fix that), we could still float it I'm sure by changing the vim.cmd to an edit instead of vsplit

Thanks. the float is a bit less intuitive for editing files. we better have a real file like `kubectl edit` does.

mosheavni:
Added a conditional for when there's no namespace.
c7e9d13ee41018f40158d18a8d14f490379027a7

Ramilito:
I'm not super into the vsplit way and floating buffer doesn't seem to play well either (too much manual work like attaching lsp and copying buffer config to mimic a normal buffer)

Went back to the original plan, we open a new tab and use a full view buffer, I think that solves your initial issues with exiting the buffer without closing the plugin.

Other minor changes:
- We fail hard if no temp file was created since that's not recoverable
- changed how we track modified to what I felt was a bit simpler (we track file modified time instead), only ugly thing was the defer_fn to allow for :wq to go through before checking.


What do you think?

mosheavni:
oh my god better than I could have achieved. I love it!! let's merge it fast 🤩

PR #225, opened 2024-08-24
feat/deployment-go-to-pods

### LINKED ISSUE #223: feat: clicking on deployment shows relevant pods
When on deployments or daemonsets page, clicking <kbd>\<enter></kbd> should show me pod view with the specific workload pods I chose before.

I'm not sure if that's easy since we have filter but filter is not quite right for this.
and using `ownerReference` field is not straight forward because the direct owner of pods is a replicaset and not a deployment.

mosheavni:
fixes #223
While it works, I don't really know based on what to switch back to all pods view.
I need some ideas on how to attack this

### DIFF
--- a/lua/kubectl/views/init.lua
+++ b/lua/kubectl/views/init.lua
@@ -1,9 +1,11 @@
 local ResourceBuilder = require("kubectl.resourcebuilder")
+local encode = require("kubectl.utils.url").encode
 local buffers = require("kubectl.actions.buffers")
 local commands = require("kubectl.actions.commands")
 local completion = require("kubectl.utils.completion")
 local definition = require("kubectl.views.definition")
 local hl = require("kubectl.actions.highlight")
+local pod_view = require("kubectl.views.pods")
 local tables = require("kubectl.utils.tables")
 
 local M = {}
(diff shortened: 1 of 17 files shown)


mosheavni:
ok i just reassigned `current_url` and only when you refresh with `gr` the view switches back to all pods

mosheavni:
maybe now move it to somewhere else and implement also on daemonsets?

mosheavni:
@Ramilito  this one is ready for review

lua/kubectl/views/init.lua
```diff
@@ -167,6 +167,39 @@ function M.UserCmd(args)
   end)
 end
 
+function M.set_and_open_pod_selector(kind, name, ns)
+  local pod_view = require("kubectl.views.pods")
+  if not kind or not name or not ns then
+    return pod_view.View()
+  end
+  local get_selectors = { "get", kind, name, "-n", ns, "-o", 'jsonpath="{.spec.selector.matchLabels}"' }
```
Ramilito, line 175: A lot of magic going on here, it's probably super correct and all but what about doing a `-o json` then 

```
local pods = vim.decode.json(object)
local labels = {}
for pod in pods then
table.insert(labels, pod.selector.matchLabels
end
```
Wouldn't that save you the trouble of handling the jsonpath output?

mosheavni: ok changed the code a bit. also found out about `vim.uri_encode()` which eliminates the need for a custom function :)))

Ramilito:
Nice! 🚀

PR #226, opened 2024-08-24
Feat/add jobs and cronjobs

mosheavni:
Edit:
fixed the issue with the flickering.
Also did:
- [X] use a mutual function for `:Kubectl get ...` and alias view: `require('kubectl.views').view_or_fallback`
- [X] restore `cmd` on fallback view. If there are no cached resources, it would use `curl` instead of `kubectl`
- [X] if resource age is more than a year, add the days difference

Original:
I want to add jobs and cronjobs for these reasons:
- [ ] be able to navigate from cronjob to job, and from job to pods
- [ ] have some nice highlights
- [ ] create job from job (`kubectl create job --from=job/...`) or cronjob (`kubectl create job --from=cronjob/...`)

For some reason, if I open `cronjobs` then go to `jobs`, the new view would flicker for less than a second and we're back to `cronjobs`!
no matter how I tried to debug it, I don't understand why I keep going back to cronjobs
Worth mentioning that if I open `jobs` before `cronjobs` the view works just fine, until I open `cronjobs` and try to go back to `jobs`.
I'm baffled here. what could cause it?

### DIFF
--- a/ftplugin/k8s_jobs.lua
+++ b/ftplugin/k8s_jobs.lua
@@ -1,0 +1,71 @@
+local api = vim.api
+local commands = require("kubectl.actions.commands")
+local definition = require("kubectl.views.jobs.definition")
+local job_view = require("kubectl.views.jobs")
+local loop = require("kubectl.utils.loop")
+local root_view = require("kubectl.views.root")
+local view = require("kubectl.views")
+
+--- Set key mappings for the buffer
+local function set_keymaps(bufnr)
+  api.nvim_buf_set_keymap(bufnr, "n", "g?", "", {
+    noremap = true,
+    silent = true,
+    desc = "Help",
+    callback = function()
+      view.Hints(definition.hints)
+    end,
+  })
+
+  api.nvim_buf_set_keymap(bufnr, "n", "<CR>", "", {
+    noremap = true,
+    silent = true,
+    desc = "Go to pods",
+    callback = function()
+      local name, ns = job_view.getCurrentSelection()
+      view.set_and_open_pod_selector("jobs", name, ns)
+    end,
+  })
+
+  api.nvim_buf_set_keymap(bufnr, "n", "<bs>", "", {
+    noremap = true,
+    silent = true,
+    desc = "Go up",
+    callback = function()
+      root_view.View()
+    end,
+  })
+
+  api.nvim_buf_set_keymap(bufnr, "n", "grr", "", {
+    noremap = true,
+    silent = true,
+    desc = "Create job from job",
+    callback = function()
+      local name, ns = job_view.getCurrentSelection()
+      vim.ui.input({ prompt = "New job name " }, function(input)
+        if not input or input == "" then
+          return
+        end
+        commands.shell_command_async(
+          "kubectl",
+          { "create", "job", input, "--from", "jobs/" .. name, "-n", ns },
+          function(response)
+            vim.schedule(function()
+              vim.notify(response)
+            end)
+          end
+        )
+      end)
+    end,
+  })
+end
+
+--- Initialize the module
+local function init()
+  set_keymaps(0)
+  if not loop.is_running() then
+    loop.start_loop(job_view.Draw)
+  end
+end
+
+init()
--- a/lua/kubectl/views/init.lua
+++ b/lua/kubectl/views/init.lua
@@ -3,6 +3,7 @@ local buffers = require("kubectl.actions.buffers")
 local commands = require("kubectl.actions.commands")
 local completion = require("kubectl.utils.completion")
 local definition = require("kubectl.views.definition")
+local find = require("kubectl.utils.find")
 local hl = require("kubectl.actions.highlight")
 local tables = require("kubectl.utils.tables")
 
@@ -164,6 +165,29 @@ function M.UserCmd(args)
       self:display("k8s_usercmd", "UserCmd")
     end)
   end)
+end
+
+function M.view_or_fallback(view_name)
+  local supported_view = nil
+  local viewsTable = require("kubectl.utils.viewsTable")
+  for k, v in pairs(viewsTable) do
+    local found_in_table = find.is_in_table(v, view_name, true)
+    if found_in_table or k == view_name then
+      supported_view = k
+    end
+  end
+  local view_to_find = supported_view or view_name
+  local ok, view = pcall(require, "kubectl.views." .. view_to_find)
+  if ok then
+    vim.schedule(function()
+      pcall(view.View)
+    end)
+  else
+    vim.schedule(function()
+      local fallback = require("kubectl.views.fallback")
+      fallback.View(nil, view_name)
+    end)
+  end
 end
 
 return M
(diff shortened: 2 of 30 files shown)


mosheavni:
I tried now `main` branch and even with the fallback view, there's problem in `jobs` view. I cannot sort on any column

mosheavni:
`set_and_open_pod_selector` depends on #225 
the problem for some reason was with `definition.display_name`

ftplugin/k8s_jobs.lua
```diff
@@ -0,0 +1,71 @@
...
+  api.nvim_buf_set_keymap(bufnr, "n", "<bs>", "", {
+    noremap = true,
+    silent = true,
+    desc = "Go up",
+    callback = function()
+      root_view.View()
+    end,
+  })
+
+  api.nvim_buf_set_keymap(bufnr, "n", "grr", "", {
```
Ramilito, line 39: Should we use a different keybind?
Grr made sense for rollout restart but not so much here

lua/kubectl/views/init.lua
```diff
@@ -166,4 +167,27 @@ function M.UserCmd(args)
   end)
 end
 
+function M.view_or_fallback(view_name)
```
Ramilito, line 170: Nice! 👍

lua/kubectl/views/jobs/definition.lua
```diff
@@ -0,0 +1,99 @@
...
+    "AGE",
+    "CONTAINERS",
+    "IMAGES",
+    "SELECTOR",
+  }
+
+  return headers
+end
+
+function M.getDuration(row)
```
Ramilito, line 56: In general, these are internal functions inside the module, we shouldn't export them unless needed ofc, not a big deal though

lua/kubectl/views/jobs/init.lua
```diff
@@ -0,0 +1,80 @@
...
+          informer.start(builder)
+          vim.schedule(function()
+            M.Draw(cancellationToken, from_cronjob)
+          end)
+        end)
+  end
+end
+
+function M.Draw(cancellationToken, from_cronjob)
+  M.builder = M.builder
```
Ramilito, line 39: Can we use state.filter  to do this instead?

mosheavni:
That's so weird! no! it's happening again.
If I change the display name to something that doesn't contain `jobs` like `CronMobs` it's working great, once I change to any variation:
`CronJobs`
`Cronjobs`
`Cron-Jobs`
or `Cron-jobs`
it flashes `jobs` quickly and then sends me right back to `cronjobs`

mosheavni:
> That's so weird! no! it's happening again. If I change the display name to something that doesn't contain `jobs` like `CronMobs` it's working great, once I change to any variation: `CronJobs` `Cronjobs` `Cron-Jobs` or `Cron-jobs` it flashes `jobs` quickly and then sends me right back to `cronjobs`

I think it might be related to something more general like some sort of globbing we have somewhere maybe?

mosheavni:
works with `Cron Jobs`. that's so weird.

mosheavni:
now it's not working again. I don't know anymore.

Ramilito:
Could you try cleaning the shada files?? I'm just guessing now

mosheavni:
> Could you try cleaning the shada files?? I'm just guessing now

Same:
```
> echo '2verbose wshada' | nvim -es ; echo
Writing ShaDa file "/Users/mavni/.local/state/nvim/shada/main.shada"
> rm -f /Users/mavni/.local/state/nvim/shada/main.shada
```

can you try locally?
open vim
open kubectl
open cronjobs
click on a cronjob
if the view flickers or stays on `cronjob` you reproduced the issue.

ftplugin/k8s_jobs.lua
mosheavni: yeah i just had no clue what to use. suggestions?

lua/kubectl/views/jobs/definition.lua
mosheavni: Done, made them local and hoisted them up.
f97cd4a..e8a63f0

lua/kubectl/views/jobs/init.lua
mosheavni: not really cause here I filter based on `ownerReference` not just any field.

Ramilito:
I think it's really awesome and the things that are left (keymap grr and the weirdness) can be separate issues!

lua/kubectl/views/jobs/init.lua
```diff
@@ -0,0 +1,80 @@
...
+local informer = require("kubectl.actions.informer")
+local tables = require("kubectl.utils.tables")
+
+local M = {
+  builder = nil,
+}
+
+function M.View(cancellationToken, from_cronjob)
+  if M.builder then
+    M.builder = M.builder
```
Ramilito, line 14: We can probably rewrite this a bit cleaner using our Resourcebuilder.view, not sure if I'm missing something though?
```
  if M.builder then
    M.builder = M.builder:view(definition, cancellationToken)
  else
    M.builder = ResourceBuilder:new(definition.resource):view(definition, cancellationToken)
  end

```

lua/kubectl/views/jobs/init.lua
Ramilito: That's true, a job can have it's own name not tied to cronjob, looking awesome!

lua/kubectl/views/jobs/init.lua
```diff
@@ -0,0 +1,80 @@
...
+
+  M.builder = M.builder
+      :process(definition.processRow):sort():prettyPrint(definition.getHeaders)
+
+  M.builder = M.builder:addHints(definition.hints, true, true, true):setContent(cancellationToken)
+end
+
+function M.Desc(name, ns)
+  ResourceBuilder:new("desc")
+      :displayFloat("k8s_job_desc", name, "yaml")
```
Ramilito, line 59: we have the Resourcebuilder.view_float helper if we aren't doing anything special here

ftplugin/k8s_jobs.lua
Ramilito: k9s seems to call it trigger, that's not that bad, also gc (create) could be an option?

lua/kubectl/views/jobs/init.lua
mosheavni: I override the call to self.draw because our draw accepts another parameter (`from_cronjob`).
maybe that's not the smartest way to do it, would love your thoughts

lua/kubectl/views/jobs/init.lua
mosheavni: I just copied it from every other view:
https://github.com/mosheavni/kubectl.nvim/blob/25f7daafe10497d756fa12d9f11c6a27b8e6be3e/lua/kubectl/views/crds/init.lua#L30-L37
https://github.com/mosheavni/kubectl.nvim/blob/25f7daafe10497d756fa12d9f11c6a27b8e6be3e/lua/kubectl/views/configmaps/init.lua#L32-L39

and many more

i think, if you want to change it, we can create an issue for changing it to EVERYTHING, not just this one.

Ramilito: I think both approaches are valid,  i.e. all new code uses the new way  vs doing a big refactor at once but no let's merge this sucker!!

mosheavni:
The display_name is still `CronMobs`, I really don't mind it as long as it works but we'll need to change it some time

PR #230, opened 2024-08-25
feat/view-or-fallback

mosheavni:
Create a reusable function to either open the requested view if exists or fallback.
handles all cases


Ramilito:
LGTM 🚀

PR #231, opened 2024-08-25
feat/add-days-to-years

mosheavni:
I have a resource that's 1 year and 11 months and 2 days
and I see that its age is `1y`
very confusing.
adding days if we specify years

### DIFF
--- a/lua/kubectl/utils/time.lua
+++ b/lua/kubectl/utils/time.lua
@@ -30,7 +30,7 @@ function M.since(timestamp, fresh, currentTime)
 
   local status = { symbol = "", value = "", sort_by = tonumber(parsed_time) }
   if days > 365 then
-    status.value = string.format("%dy", years)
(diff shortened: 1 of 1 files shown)


Ramilito:
I see, wouldn't months be better though?

mosheavni:
> I see, wouldn't months be better though?

I think it would but not all months are same number of days :)

Ramilito:
I feel 500+ days is crazy but I also don't like 1y and 200 days, we used to have it as y/m/d, should we revert to that?

Maybe k9s has a better way?

mosheavni:
Just checked k9s and they do year days
Just like jn this PR
I'll post a screenshot soon

mosheavni:
image
they add days only if it's more than 2 years.
I don't see why not adding days over 1 year.

Ramilito:
I used ti have days here and removed it, needed to be convinced to go back on that 😅 thanks for your patience though 🙏

PR #233, opened 2024-08-25
Feat/session per context

### LINKED ISSUE #227: feat: save session per context
right now when we save the session to `kubectl.json`, we don't save the kube-context, so whenever we switch context, the view is reset to `pods` and without namespace.

I would like that when there's no kube-context record on `kubectl.json`, we would load the `namespace` and `view` from the current one, and if there's a record, use that session details

mosheavni:
fixes #227


mosheavni:
it's working well but it requires to delete `kubectl.json` before.
Do we want to handle it in the code or just leave it for the user to delete the file?

Ramilito:
I think we should try to have the habit of not disrupting peoples work, how about adding a "try/catch" in our restore session function that if it fails, it also deletes the file?

We can make it smart in trying to save some of the information and recreating the file with history if we want later but at a minimum we should solve the file incompatibility issue

lua/kubectl/state.lua
```diff
@@ -121,18 +121,13 @@ function M.restore_session()
   end
 
   -- Restore state
-  M.ns = M.session.namespace
+  local context_table = M.session.contexts[M.getContext().contexts[1].name] or M.session.contexts.default
```
Ramilito, line 124: I'd love some if checks on that get context, maybe in the getcontext it self where it returns a default value so we don't get any raceconditions when trying to retrieve that information..

mosheavni: honestly would appreciate some help here 🙌🏽

mosheavni:
@Ramilito  I could really use this feature 🙏🏼

Ramilito:
On it!

Ramilito:
I think this is working, could you test it out?

mosheavni:
Not all of the contexts are saved on first time.
- open `k8s` on a contexts that is not saved.
- switch to a different context
- exit vim

what happens is that only the second context (or probably the last one) is being saved to the config

Ramilito:
It's supposed to only care about the current context. Is it at least doing that?

My plan was to check if you have a current context in your session file, if not, load defaults and create it when exiting.
Next time you open the plugin while in the same context, it should have rememberd, did that not work at all?

mosheavni:
Yes it does keep the last one
I guess I can live with that, only a matter of populating the config with all of my context once, and forget about it.
Thanks for your work here
Ok to merge by me 🙏🏼

Ramilito:
Yeah I actually forgot that you mentioned we should save the session when running Kubectx as well and now it's super late, maybe we can do something later if it bothers you 👍

PR #234, opened 2024-08-25
feat/wrap-log-lines

mosheavni:
Allow toggling wrap on log lines


Ramilito:
Should we use the term wrap? with a keybinding of gw?
I was a bit confused on what collapse was before seeing the code :D

mosheavni:
Sorry I had issues with `gw` but their gone.
changed :)

Ramilito:
🚀

PR #235, opened 2024-08-25
fix/describe-on-fallback-cluster-scoped

mosheavni:
describe and edit on fallback resources only works for namespaced items
this feature branch defines another `definition` value for fallback.definition: `M.namespaced`
that determines how `getCurrentSelection` gets the name / namespace+name of the resource


lua/kubectl/views/fallback/definition.lua
```diff
@@ -109,6 +110,8 @@ function M.getHeaders(rows)
     local firstRow = rows.rows[1]
     if firstRow and firstRow.object and firstRow.object.metadata and firstRow.object.metadata.namespace then
       table.insert(headers, "NAMESPACE")
+    else
+      M.namespaced = false
```
Ramilito, line 114: You think we can use this information instead?

We already know if a resource is namespaced or not by your other code in https://github.com/Ramilito/kubectl.nvim/blob/41333f68718e6ae4e26d67716cd091f01a854b53/lua/kubectl/views/definition.lua#L109

So if we add that into the cached_api_resource object, maybe we can simplify this code a bit

mosheavni: you are right.
I did what you asked, though I'm not sure it's enough, can you please check again and let me know if there's anything you'd change? @Ramilito

Ramilito: Il take a look soon 👍

lua/kubectl/views/fallback/init.lua
```diff
@@ -77,7 +84,10 @@ end
 --- Get current seletion for view
 ---@return string|nil
 function M.getCurrentSelection()
-  return tables.getCurrentSelection(2, 1)
+  if definition.namespaced then
```
Ramilito, line 87: This would only work for the CRDS that have that variable right?

mosheavni: no, I added it to every resource here:
https://github.com/Ramilito/kubectl.nvim/pull/235/files#diff-4f77b793ea3926dba9cc1fa057d7e215c1ba56312288dca56a4519c1e903066bR116

Ramilito:
Hmm I can't find anything wrong here, is there something not working properly?

mosheavni:
> Hmm I can't find anything wrong here, is there something not working properly?

LGTM ok to merge :)

PR #237, opened 2024-08-25
feat/fallback-for-vim-nil

mosheavni:
For some reason, the fallback view for `LimitRanges` is returning `vim.NIL` on `rows.rows` if there are no objects.

This makes sure we won't fall on vim.NIL in fallback again.


Ramilito:
It's how vim.json.decode works, shoud we tell it to stop with the vim.NIl stuff instead?
image

Ramilito:
Looking awesome 👏

PR #238, opened 2024-08-25
feat/capitalize-events

mosheavni:
I have custom resources: kibanas and elasticsearches
they both have `green`, `yellow` or `red` on their status.
not capitalized.
I figured we better capitalize anyway to get the colors right.
before:
image

after:
image

### DIFF
--- a/lua/kubectl/utils/events.lua
+++ b/lua/kubectl/utils/events.lua
@@ -106,7 +106,7 @@ function M.ColorStatus(status)
   if type(status) ~= "string" then
     return ""
   end
-  local capitalized = string_util.capitalize(status)
+  local capitalized = string_utils.capitalize(status)
   if errorStatuses[capitalized] then
     return hl.symbols.error
   elseif warningStatuses[capitalized] then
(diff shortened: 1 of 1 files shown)


lua/kubectl/utils/events.lua
```diff
@@ -86,6 +88,7 @@ function M.ColorStatus(status)
     Normal = true,
     VolumeResizeSuccessful = true,
     FileSystemResizeSuccessful = true,
+    SecretSynced = true,
```
mosheavni, line 91:
```suggestion
```

lua/kubectl/utils/events.lua
```diff
@@ -95,17 +98,22 @@ function M.ColorStatus(status)
     -- Custom statuses
     Active = true,
     True = true,
+    SuccessfullyReconciled = true,
```
mosheavni, line 101:
```suggestion
    SecretSynced = true,
    SuccessfullyReconciled = true,
```

Ramilito:
There is a capitalization method in the utils/string.lua, do we do something different here?

mosheavni:
didn't know about it. now it's using it

Ramilito:
Love it!

Ramilito:
> didn't know about it. now it's using it

Yeah figured but my git diff on phone is sometimes hard to read 😅

PR #241, opened 2024-08-25
feat/sort-by-ready

mosheavni:
I asked bing-chat:
image

Here is the after and before (excuse me i placed them wrong, i'm tired probably):
image


Ramilito:
That's clever

PR #250, opened 2024-08-26
Feat/change url and show view

mosheavni:
move some code around so it's reusable


mosheavni:
it's working when the definition.cmd is `curl`
it's not working when opening the plugin on a fallback resource and the cmd is `kubectl` even though I add `-l app=my-app` to the definition.args

any idea why not? @Ramilito

Ramilito (approved):

Ramilito:
Did you mean to make this comment in the other pr? The label filter one?

mosheavni:
> Did you mean to make this comment in the other pr? The label filter one?

yes sorry.

PR #251, opened 2024-08-26
feat/pvc-to-pv

mosheavni:
click on PVC to go to the corresponding PV

I decided it's best to use a filter here since we're not going to get the same result twice
Plus I really don't see why not use a jsonpath here, does it really matter in this specific case?


Ramilito:
If we don't need to parse it then it's all good!

Ramilito:
Looks great! Ready to merge?

mosheavni:
Yes @Ramilito

PR #254, opened 2024-08-26
feat/namespace-view-or-switch

mosheavni:
If you type: `:Kubens` without args, open the namespace view
If you type: `:Kubens some-ns`, change to it

remove the duplicated function to change NS since we already have it in the namespace view

future thoughts:
do the same for `:Kubectx`

### DIFF
--- a/lua/kubectl/completion.lua
+++ b/lua/kubectl/completion.lua
@@ -91,16 +91,7 @@ function M.list_namespace()
       table.insert(ns, namespace)
     end
   end
-
   return ns
-end
-
---- Change namespace
---- @param cmd string
-function M.change_namespace(cmd)
-  local results = commands.shell_command("kubectl", { "config", "set-context", "--current", "--namespace=" .. cmd })
-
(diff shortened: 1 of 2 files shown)


mosheavni:
with this: d764949
I improve the `:Kubens ` completion by saving namespaces in the namespace view.

lua/kubectl/views/namespace/init.lua
```diff
@@ -44,7 +47,30 @@ function M.View()
...
+    end
+  end
+
+  M.namespaces = ns
+  return M.namespaces
+end
+
 function M.changeNamespace(name)
+  -- don't change NS if we know it's not valid
+  if not (#M.namespaces > 1 and vim.tbl_contains(M.namespaces, name)) then
```
Ramilito, line 71: That could cause issues for the people that type namespaces that they are not allowed to list

mosheavni: ok removed it for now.

Ramilito:
Nice!

PR #255, opened 2024-08-26
fix: use exact match for checking buffer

Ramilito:
Apparently vim.fn.bufname is not using exact matching....


mosheavni:
haha.. makes sense now.

Ramilito:
haha yeah, but you predicted it, wp!

PR #256, opened 2024-08-26
Feat/move error statuses

mosheavni:
Edit:
I added a lot of statuses I got from my clusters and sorted them.

image


Ramilito:
nice!

PR #258, opened 2024-08-26
fix: job view params

Ramilito:
Refactor so we don't need to send params into the draw function
Added filter helper

Need to reset the filter from job view


mosheavni:
nice!

mosheavni:
image

Ramilito:
huh, that's weird, are there jobs without ownerreferences...I guess you can create a job directly, I'll add a nil check!

Ramilito:
Should work better now!

PR #259, opened 2024-08-26
fix/no-containers-on-filter

mosheavni:
there's an issue when I hit <kbd>\<enter></kbd> on a pod when there is a filter going on, we won't see any container.

I checked who is using `ResourceBuilder:view_float` and it's only pods and containers.
in both cases we want to pass `no_filter` as true to `:process()`
and `include_context` and `include_filter` both as `false` to `:addHints()`


Ramilito:
Nice catch

Issue #262, opened 2024-08-26
flickering colors


mosheavni:
**Describe the bug**
when moving the cursor, it seems that the highlights are getting reset and that looks like flickering

![2024-08-27_00-18-40 (1)](https://github.com/user-attachments/assets/8cebb556-6c25-40e2-bf86-933b71882292)
(the gif kinda slows it down, it's more pronounced in actual usage)

**To Reproduce**

2. Press '....'

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. OSX]
 - Neovim version [e.g. 0.10]

**Additional context**

Ramilito:
This is fine now right?

PR #263, opened 2024-08-26
fix/flicker-on-extmarks

mosheavni:
it seems it resolves the flickering of #262 
I wonder, was there a specific reason for scheduling this block? @Ramilito


Ramilito:
hmm I can't remember, there are two reasons to use that block, one is if we are inside an async function, the other one is performance. 

I just checked and having 600+ pods does hang the ui if that isn't in a sheduling block...let me think a bit about what we can do about it.

PR #264, opened 2024-08-27
fix: flickering colors

Ramilito:
Seems to be a bad idea to schedule extmarks in favor of faster data to the reader since it will cause flickering in colors as mentioned #262.

Bonus:
- Noticed that pretty_print was taking ~10ms at 600+ pods, we don't really do that much work, managed to refactor it down to ~3ms.
- Moved schedule to wrap whole setcontent

### DIFF
--- a/lua/kubectl/actions/buffers.lua
+++ b/lua/kubectl/actions/buffers.lua
@@ -51,44 +51,42 @@ function M.apply_marks(bufnr, marks, header)
   api.nvim_buf_clear_namespace(bufnr, ns_id, 0, -1)
   state.marks.ns_id = ns_id
 
-  vim.schedule(function()
-    if header and header.marks then
-      for _, mark in ipairs(header.marks) do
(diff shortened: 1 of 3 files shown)


mosheavni:
Amazing
How did you calculate the benchmark of pretty print?

Ramilito:
> Amazing How did you calculate the benchmark of pretty print?

I'm using two ways, the easiest is just to use the timeme utility we have in utils/timeme.lua and surround a function with timeme.start()
.....
timeme.stop()

This just prints out the time in seconds and works great when you know where you need to improve.

The second way is how I usually start from, using the plugin: "stevearc/profile.nvim"
It gives me this view:

Before:
image

After:
image

mosheavni:
Awesome! I'll start using that on my PRs :)
thank you for the detailed explanation.

Ramilito:
I'm hoping this solves the flickering and gives us a snappier ui!

PR #268, opened 2024-08-27
Fix/replicas of deployments

mosheavni:
It's really not super intuitive understanding how they represent the lhs and rhs of READY column.

I tested on multiple clusters, compared the result to `kubectl get deployment` output.

it seems I got it right :)


mosheavni:
forgot to show the before:
image

PR #270, opened 2024-08-27
Feat/move keymaps to config

mosheavni:
start moving keymaps to config

because of the vim loop, the global keymaps are being set before the config is set up. how can we go around it?


mosheavni:
it's working
i think this PR is already long enough
we can split it and next:
- add keybindings overrides to the rest of the keymaps
- generate hints dynamically

PR #272, opened 2024-08-28
Feat/top view

mosheavni:
basic implementation of top for nodes and pods

sometimes values are 0  and 0 for like a second and then they refresh well. we can address this later.
ready to merge.

### DIFF
--- a/lua/kubectl/actions/informer.lua
+++ b/lua/kubectl/actions/informer.lua
@@ -159,7 +159,11 @@ end
 local function on_exit() end
 
 function M.start(builder)
-  if not builder.data or not builder.data.metadata or builder.data.kind == "PodMetricsList" then
+  local metrics = {
+    "NodeMetricsList",
+    "PodMetricsList",
+  }
+  if not builder.data or not builder.data.metadata or vim.tbl_contains(metrics, builder.data.kind) then
     return
   end
   if M.handle then
--- a/lua/kubectl/actions/informer.lua
+++ b/lua/kubectl/actions/informer.lua
@@ -159,7 +159,7 @@ end
 local function on_exit() end
 
 function M.start(builder)
-  if not builder.data or not builder.data.metadata then
+  if not builder.data or not builder.data.metadata or builder.data.kind == "PodMetricsList" then
     return
   end
   if M.handle then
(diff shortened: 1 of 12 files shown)


mosheavni:
Done!

mosheavni:
sorting not working.
i'm on it

mosheavni:
ok fixed sort, welcome to test

mosheavni:
not accumulating multi container metrics

lua/kubectl/actions/informer.lua
```diff
@@ -159,7 +159,11 @@ end
 local function on_exit() end
 
 function M.start(builder)
-  if not builder.data or not builder.data.metadata then
+  local metrics = {
```
Ramilito, line 162: I see what you are trying to do here and while it works it's kind of hidden behaviour, how about moving the idea to the Resourcebuilder:view and changing this instead:
Before:
```
      if opts.cmd == "curl" then
        informer.start(builder)
      end

```
After:
```
      if opts.cmd == "curl" and opts.with_informer then
        informer.start(builder)
      end

```

lua/kubectl/views/init.lua
```diff
@@ -100,11 +100,24 @@ function M.Hints(headers)
   buffers.set_content(buf, { content = content, marks = marks })
 end
 
+local function merge_views(cached_resources, views_table)
```
Ramilito, line 103: Great idea! Maybe we could move the function to definition.lua though?

mosheavni: Done. sheesh that was exhausting calculating the cpu and mem...

mosheavni:
i honestly think it's ready now @Ramilito

lua/kubectl/views/init.lua
Ramilito: lol yeah I was like, I'll read that part later 🤣

Ramilito:
Lets merge it! I'll create a issue to remove the informer part and add an with_informer option to the builder

lua/kubectl/actions/informer.lua
mosheavni: done.
since i don't want to add {informer = true} to every view, I just check if it's set to false explicitly

PR #273, opened 2024-08-28
feat: configurable keymaps

Ramilito:
This is working well and allows the user to override any functions they want.

Remaining is to update the hints with the correct key when overridden (should be easy to lookup) and to update readme

- [x] Add plugin prefix to our plug functions
- [ ] Update readme
- [x] Dynamic hints


Ramilito:
@mosheavni  I was thinking more along these lines instead of creating our own DSL

mosheavni:
> @mosheavni I was thinking more along these lines instead of creating our own DSL

how so?

Ramilito:
> > @mosheavni I was thinking more along these lines instead of creating our own DSL
> 
> how?

`<Plug>` is a vim keyword, the views would just expose a function and not bother with who triggers the function.
We would then see if that function is mapped to anything and if it is not we set the default ourselves.


The user would override the keymap like this example below:
2024-08-28T12-27-18Z_code

lua/kubectl/views/clusterrolebinding/definition.lua
```diff
@@ -5,7 +5,7 @@ local M = {
   ft = "k8s_clusterrolebinding",
   url = { "{{BASE}}/apis/rbac.authorization.k8s.io/v1/clusterrolebindings?pretty=false" },
   hints = {
-    { key = "<gd>", desc = "describe", long_desc = "Describe selected clusterrolebinding" },
+    { key = "<Plug>(kubectl.describe)", desc = "describe", long_desc = "Describe selected clusterrolebinding" },
```
mosheavni, line 8: do we really need to list it here?
it's part of the global maps and I don't see why see it in the header

mosheavni: i mean every definition that defines a global hint (such as describe)

Ramilito: yeah I'll move it to the global hints window!

Ramilito: also moved the g? from every ftplugin, made it work on floats or catch errors

Ramilito:
I'll merge it so you can use it, will add the leftover to "update readme" issue

PR #275, opened 2024-08-28
fix/remove-metrics-logic-from-informer

mosheavni:
fixes of #272


Ramilito:
That was fast!

Shall we merge?

mosheavni:
yeah :)

PR #276, opened 2024-08-28
Feat/toggle follow

mosheavni:
toggle follow logs

also start to consolidate container_logs and pod_logs which are pretty similar

also keymap `<cr>` to add a divider, VERY useful for whoever is following fast lines.


mosheavni:
ready @Ramilito

lua/kubectl/views/pods/init.lua
```diff
@@ -9,6 +9,8 @@ local M = {
   builder = nil,
   selection = {},
   pfs = {},
+  is_tailing = false,
+  tail_handle = nil,
```
Ramilito, line 13: I think you can improve the code a little bit byt using the handle as a check instead of having two variables, it's not much but a small improvement.
you would do something like this:
```
      if handle and not handle:is_closing() then
        handle:kill(2)
        handle = nil
      else
        handle = start_tailing
      end

```

mosheavni:
I cannot remap `<cr>` for adding a divider. ideas?

lua/kubectl/views/pods/init.lua
mosheavni: fixed.

mosheavni:
> I see you fixed it?

yes, also saved the winnr into M in order to not try and set the cursor if the tail window is closed (happened to me when i clicked `f` and closed the window very fast)

also fixed an issue where i follow a log, close it and open a different log, container would not work.
used a local var instead of `M` one.
I think this PR is ready.

lua/kubectl/views/pods/init.lua
```diff
@@ -26,40 +27,68 @@ function M.Draw(cancellationToken)
...
+      end,
+    })
+  end
 end
 
-function M.selectPod(pod_name, namespace)
-  M.selection = { pod = pod_name, ns = namespace }
+function M.selectPod(pod, ns)
+  local container
+  local data = M.builder.data.items
```
Ramilito, line 78: I'm a bit sceptical about this approach, what are we trying to do here?

mosheavni: if you follow logs for a pod, kubectl will select the first container automagically
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_logs/
according to this we can add the flag: `--all-containers=true`
so it needs to be refactored this way.
Unfortunately I don't have any time today and tomorrow i'm going on a vacation for a week.
Please feel free to step in on this branch and do whatever you think.
I think toggling follow is an important feature.
before this, I would need to leave the log view in order to stop following and sometimes I just follow logs and just want to stop and investigate the output.

every feature I add is out of real life usage of the plugin :)

Ramilito: I agree and appreciate your effort a lot ❤️, wasn't questioning your motives nor the idea!
Was trying to balance the feature against performance and sustainability and a gut feeling that we could do something better here but wanted to know if I've missed something that would make this a valid approach.
We can merge and have that as a single fix up issue ✌️

Ramilito: And also! Have a nice vacation! Well deserved IMHO!!

PR #281, opened 2024-08-29
chore: defaulting to all containers when no container is specified

Ramilito:
defaulting to all containers when none is specified
using kubectl for the initial opening of logs view since curl only supports one container per session

### DIFF
--- a/ftplugin/k8s_pod_logs.lua
+++ b/ftplugin/k8s_pod_logs.lua
@@ -4,6 +4,7 @@ local str = require("kubectl.utils.string")
 
 mappings.map_if_plug_not_set("n", "f", "<Plug>(kubectl.follow)")
 mappings.map_if_plug_not_set("n", "gw", "<Plug>(kubectl.wrap)")
+mappings.map_if_plug_not_set("n", "gp", "<Plug>(kubectl.prefix)")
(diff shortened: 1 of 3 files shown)


mosheavni:
image
view is not full when following

Ramilito:
Hmmmmm...I'm not seeing that, how did you get that results?

mosheavni:
create this pod:
```
cat <<EOF | kaf -
apiVersion: v1
kind: Pod
metadata:
  name: incrementing-pod
  annotations:
    sidecar.istio.io/inject: "false"

spec:
  containers:
  - name: incrementing-container
    image: alpine
    command: ["/bin/sh", "-c"]
    args:
      - |
        i=0
        while sleep 1; do
          echo hello $i
          i=$((i + 1))
        done

  - name: incrementing-container2
    image: alpine
    command: ["/bin/sh", "-c"]
    args:
      - |
        i=0
        while sleep 1; do
          echo bye $i
          i=$((i + 1))
        done
EOF

```

click `gl` on the pod, then `f` to follow

Ramilito:
Yeah that made sense, adjusted it

Ramilito:
What do you think? merge?

mosheavni:
checking

mosheavni:
LGTM! very nice!

mosheavni:
logs feature is 10000 times better now than it was before! amazing work.

Issue #286, opened 2024-08-29
Completion for filter


mosheavni:
**Is your feature request related to a problem? Please describe.**
Sometimes I need to use the previous filter, can we offer it as a completion like in namespace and alias menu?

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Ramilito:
done!

PR #287, opened 2024-08-29
feat: add completion to filter

Ramilito:
added completion to filter
exclude namespace from filter

solves #286


mosheavni:
very nice! any idea why it sends me to the start of the line upon completion?

Ramilito:
hmm where are you seeing that? after filtering?

Ramilito:
ah, that's probably `vim.api.nvim_win_set_cursor(0, { #header + 2, #prompt })`
We should check if cursor already is in prompt before calling that, it's supposed to help you when you select an option instead of using completion

Ramilito:
fixed in #294

PR #290, opened 2024-08-29
fix/delete-on-fallback

mosheavni:
I guess you never see a problem until you actually try to do something.
`gD` did not work on fallback view


Ramilito:
Yeah I'm starting to think we might need to have some tests for this.
Looks great! merging!

PR #302, opened 2024-09-08
feat/port-forward-keymaps

mosheavni:
add keymap `<CR>` to change container port
add keymap `gr` to reset the view if something is messed up and we wanna start over


lua/kubectl/actions/buffers.lua
```diff
@@ -163,6 +163,7 @@ end
 --- @param filetype string: The filetype of the buffer.
 --- @param onConfirm function: The function to call on confirmation.
 --- @param opts { syntax: string|nil, content: table|nil, marks: table|nil, width: number|nil }|nil: Options for the buffer.
+--- @return integer, table: The buffer and window number.
```
Ramilito, line 166: A bit pedantic but it's the window config, not number 😅

mosheavni: absolutely right.

PR #304, opened 2024-09-09
fix/pod-status-fallback

mosheavni:
sometimes I get an error for `row.status` is `nil`
this ensures we don't see these messages


Ramilito:
LGTM!

PR #317, opened 2024-09-19
fix: don't include port in the url when 443 or 80

Ramilito:
When browsing to a host that has 80 or 443 as port number we are getting http://url80


mosheavni:
sorry my bad

Issue #316, opened 2024-09-19
history on aliases view


mosheavni:
just like when filtering, I can see the last 5 filters, when on `aliases` float view, I'd like to see my last 5 resources i visited :)

Ramilito:
Great idea!

PR #318, opened 2024-09-19
feat: alias history

Ramilito:
Fixes: #316 

Adds a selectable history list, we don't bother adding the history to the tab completion since they should be there already.
A bit of copy/paste from filter but I think it's fine until we want this in more places.


mosheavni:
thank you!

PR #319, opened 2024-09-19
Fix/filter not filtering

### LINKED ISSUE #308: filter does not work when switching from pod to ingress view
**Describe the bug**
ok this is weird, I started experiencing it a while ago but disregarded it.
now it's bugging me
when opening straight to pods view, then applying a filter, then switching to ingress, even though there are resources that should match the filter, I see 0 ingresses!
I ran a `git bisect` and found that this: b637e870ba5942697314afa1aa65547a3ba35e95 is the first bad commit:

```
b637e870ba5942697314afa1aa65547a3ba35e95 is the first bad commit
```
which makes this the last good commit: a07612541e5b4fe05f5309460def2e1a0ced5f75
```
a07612541e5b4fe05f5309460def2e1a0ced5f75
```

**Bad commit reproduction**
1. make sure we're aligned to origin/maon
2. open `k8s` on `pods` view
3. filter for a value
4. switch to `ingress` view
5. filter is applied, ingress exists, yet I see 0!

**Good commit reproduction**
1. reset to this commit:
```
git checkout a07612541e5b4fe05f5309460def2e1a0ced5f75
```
2. open `k8s` on `pods` view
3. filter for a value
4. switch to `ingress` view
5. filter is applied, ingress exists, i see the ingress

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. OSX]
 - Neovim version [e.g. 0.10]

**Additional context**

mosheavni:
fixes #308


mosheavni:
it's because when we're on fallback view, the `name` is a table rather than a string (on defined views it's always a string):
https://github.com/mosheavni/kubectl.nvim/blob/8d664f366ef9afb3d842ffa3ec33805ba640d9f3/lua/kubectl/views/fallback/definition.lua#L73

then when we need to recuresively drop into `name`, it's using exact match which is almost never what we want, we want the pattern to be a substring of the field

Ramilito:
Nice catch, do you have any idea why I couldn't replicate the bug??

mosheavni:
I think if you search for a substring of some field in the fallback view, you wouldn't see it, only if it's exactly the same name

PR #320, opened 2024-09-19
fix/dont-delete-existing-float

mosheavni:
when the float buffer already exist, don't set the content to `Loading...`


Ramilito:
This is included in #322, did a small difference but it was inspired by this PR

PR #326, opened 2024-09-21
feat: enhance kubectl apply

Ramilito:
- [x] Show diff of changes
- [x] Show context

image

### DIFF
--- a/lua/kubectl/completion.lua
+++ b/lua/kubectl/completion.lua
@@ -153,6 +153,9 @@ function M.apply()
         end
       end)
 
+      if #builder.data == 1 then
+        table.insert(builder.data, "[Info]: No changes found when running diff.")
+      end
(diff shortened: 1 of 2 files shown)


mosheavni:
i'm not sure how to use it.
I open a file, change some stuff and run `:Kubectl apply` and it gives me:
image

I did not start the plugin before that.

do we want to:
- Initiate the plugin wherever someone uses `:Kubectl` or `:Kubens` or `:Kubectx`?
- if the plugin is not initialized, do nothing.

Leaving it as is, where it produces an error is not great IMO

Ramilito:
Removed the current-context prompt, could you see if it handles better? It should work without starting the plugin this way

mosheavni:
> Removed the current-context prompt, could you see if it handles better? It should work without starting the plugin this way

yep I can confirm it's working now

PR #327, opened 2024-09-21
Feat/filter visual selection

mosheavni:
add keymap for filtering based on the current visual selection.
I couldn't believe this is so simple.. spent far too much time on it.
feel free to move the function anywhere else, I had no better idea.


lua/kubectl/utils/string.lua
```diff
@@ -21,4 +21,12 @@ function M.divider(buf, char)
   vim.api.nvim_win_set_cursor(0, { line_count + 1, 0 })
 end
 
+function M.get_visual_selection()
+  local esc = vim.api.nvim_replace_termcodes("<esc>", true, false, true)
```
Ramilito, line 25: I dislike how vim gets visual selection so much...

Ramilito:
It's perfect! really awesome addition!

lua/kubectl/utils/string.lua
mosheavni: Yes trust me, currently, there is no other way..

PR #328, opened 2024-09-22
chore: global keymaps buf only

mosheavni:
when I use the plugin, it's either I opened vim and went into the plugin, or I have to exit and then work regularly on nvim.
that is because of the mappings.
I have <kbd>\<C-f></kbd> mapped to something else which I use all the time.
I have <kbd>\<CR></kbd> mapped to `:noh<cr><cr>`
many more examples

Also, when I edit a resource (<kbd>ge</kbd>) I need my real keymaps.

to sum up, I don't see any reason why not to map the global plugin keymaps to be buffer local


Ramilito:
That's a good point! If it's ready lets merge it!

mosheavni:
works well for me :)

PR #329, opened 2024-09-22
chore/fix-nil-status-pod-ip

mosheavni:
another fallback.
there are many cases like this.. gotta catch'em all
image


Ramilito:
It's like pokemons, we can reflect on these and see if there is a better way to handle exceptions, but for now lets do it like this 👍

PR #330, opened 2024-09-22
fix/plug-keymaps-override

mosheavni:
My last PR #328  messed up the `Plug` keymap overrides.
This fixes it.


Ramilito:
Right, that's reasonable

mosheavni:
> Right, that's reasonable, nice quick catch

thanks.. I'll try to remember to test more before I merge.

Ramilito:
It's ok, we are doing awesome stuff here, really happy with how this plugin is turning out and a lot of it is thanks to you ❤️

PR #332, opened 2024-09-22
fix/secret-desc-error-and-hint

mosheavni:
fix base64 decode errors that we don't wanna see and display the right Hint for secret describe float


Ramilito:
LGTM!

PR #335, opened 2024-09-22
fix/deployment-status-nil

mosheavni:
found another nil status.
@Ramilito  we must think of a more general solution. WDYT?


Ramilito:
I agree but let's get this one in, not sure what the best approach is

PR #336, opened 2024-09-22
chore/get-external-ip-for-svc

mosheavni:
perhaps there are other things that give this field the value but for my clusters, all I could find is `status.loadBalancer.ingress[1].hostname`

### DIFF
--- a/lua/kubectl/views/services/definition.lua
+++ b/lua/kubectl/views/services/definition.lua
@@ -36,6 +36,14 @@ local function getType(type)
   return { symbol = typeColor[type] or "", value = type }
 end
 
+local function getClusterIP(row)
+  local clusterIP = row.spec.clusterIP or "<none>"
+  if clusterIP == "None" then
+    clusterIP = "<none>"
+  end
+  return clusterIP
+end
+
 local function getExternalIP(row)
   return row.spec.externalName
     or row.status
@@ -58,7 +66,7 @@ function M.processRow(rows)
       namespace = row.metadata.namespace,
       name = row.metadata.name,
       type = getType(row.spec.type),
-      ["cluster-ip"] = row.spec.clusterIP or "<none>",
+      ["cluster-ip"] = getClusterIP(row),
       ["external-ip"] = getExternalIP(row),
       ports = getPorts(row.spec.ports),
       age = time.since(row.metadata.creationTimestamp),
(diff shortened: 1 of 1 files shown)


Ramilito:
This is how the ingresses work, could it be the same?
```
local function get_address(row)
  local addresses = {}
  if row.status and row.status.loadBalancer and row.status.loadBalancer.ingress then
    for _, ingress in ipairs(row.status.loadBalancer.ingress) do
      if ingress.hostname then
        table.insert(addresses, ingress.hostname)
      elseif ingress.ip then
        table.insert(addresses, ingress.ip)
      end
    end
  end

  return table.concat(addresses, ", ")
end

```

mosheavni:
> This is how the ingresses work, could it be the same?
> 
> ```
> local function get_address(row)
>   local addresses = {}
>   if row.status and row.status.loadBalancer and row.status.loadBalancer.ingress then
>     for _, ingress in ipairs(row.status.loadBalancer.ingress) do
>       if ingress.hostname then
>         table.insert(addresses, ingress.hostname)
>       elseif ingress.ip then
>         table.insert(addresses, ingress.ip)
>       end
>     end
>   end
> 
>   return table.concat(addresses, ", ")
> end
> ```

I'm not familiar with a case where a `kind: Service` can have multiple addresses but maybe I'm wrong

Ramilito:
Tbh I'm not sure I nailed it in the ingresses resource either but took a look at k9s implementation and it looks like this:

```
func getSvcExtIPS(svc *v1.Service) []string {
	results := []string{}

	switch svc.Spec.Type {
	case v1.ServiceTypeClusterIP:
		fallthrough
	case v1.ServiceTypeNodePort:
		return svc.Spec.ExternalIPs
	case v1.ServiceTypeLoadBalancer:
		lbIps := lbIngressIP(svc.Status.LoadBalancer)
		if len(svc.Spec.ExternalIPs) > 0 {
			if len(lbIps) > 0 {
				results = append(results, lbIps)
			}
			return append(results, svc.Spec.ExternalIPs...)
		}
		if len(lbIps) > 0 {
			results = append(results, lbIps)
		}
	case v1.ServiceTypeExternalName:
		results = append(results, svc.Spec.ExternalName)
	}

	return results
}
```

mosheavni:
ok gave it ago. it's working for me but there are cases here which I cannot test.

- even on NodePort services, I don't have the field `externalIPs`
- LB ingress on my clusters is always a hostname and never an IP.

Ramilito:
That enough for me, if someone has issues they can at least see how to solve it!

🚀

PR #346, opened 2024-09-23
Feat/api resources view

### LINKED ISSUE #343: Create a view for api-resources
much like alias menu on k9s
display the output of `kubectl api-resources` and allow clicking on a resource to see it
we can maybe use the cached_resources we already have

mosheavni:
View for api-resources fixes #343 
image

### DIFF
--- a/ftplugin/k8s_api_resources.lua
+++ b/ftplugin/k8s_api_resources.lua
@@ -29,7 +29,7 @@ end
 local function init()
   set_keymaps(0)
   if not loop.is_running() then
-    loop.start_loop(api_resources_view.View)
+    loop.start_loop(api_resources_view.Draw)
   end
 end
 
--- a/ftplugin/k8s_api_resources.lua
+++ b/ftplugin/k8s_api_resources.lua
@@ -1,0 +1,36 @@
+local api = vim.api
+local api_resources_view = require("kubectl.views.api-resources")
+local loop = require("kubectl.utils.loop")
+local overview_view = require("kubectl.views.overview")
+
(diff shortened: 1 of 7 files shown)


lua/kubectl/views/api-resources/init.lua
```diff
@@ -0,0 +1,40 @@
...
+function M.View(cancellationToken)
+  local self = state.instance
+  if not self or not self.resource or self.resource ~= definition.resource then
+    self = ResourceBuilder:new(definition.resource)
+  end
+
+  self:display(definition.ft, definition.resource, cancellationToken)
+
+  state.instance = self
+  self.data = cached_resources and cached_resources.values or {}
```
Ramilito, line 17: cached_resources doesn't seem to be defined here ☝️

mosheavni: oops. fixed

ftplugin/k8s_api_resources.lua
```diff
@@ -0,0 +1,36 @@
...
+      view.view_or_fallback(name)
+    end,
+  })
+end
+
+--- Initialize the module
+local function init()
+  set_keymaps(0)
+  if not loop.is_running() then
+    loop.start_loop(api_resources_view.View)
```
Ramilito, line 32: Do we want to rerun view or draw in this loop?

mosheavni: yeah why not. changed to `Draw`

Ramilito (approved):

PR #348, opened 2024-09-23
feat/go-up-history

mosheavni:
Maintain a history of visited views, then allow to go up using backspace until we reach the end ~which is `overview` view~.
(overview is crashing my vim constantly)

I like it better than randomly going up to overview...
WDYT? @Ramilito

### DIFF
--- a/ftplugin/k8s_configmaps.lua
+++ b/ftplugin/k8s_configmaps.lua
@@ -1,19 +1,8 @@
-local loop = require("kubectl.utils.loop")
-local overview_view = require("kubectl.views.overview")
-local api = vim.api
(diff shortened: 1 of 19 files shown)


Ramilito:
I've thought about it before and agree, the initial idea of navigating in hierarchy wasn't all that great.
Let me know when you feel it's ready to merge.

I'll create an issue for the overview, didn't know it wasn't working 🥺

mosheavni:
Yes it's ready. right now I'm not saving it into the `kubectl.json` file, I don't really see what for but let me know if you think otherwise.

Ramilito:
No I agree, don't see the point in saving it either

Ramilito (approved):

PR #351, opened 2024-09-23
fix/reload-api-resources-on-ctx

mosheavni:
fixes PR-347
resources needs to be cleared and we need to ignore timestamp


Ramilito:
LGTM!
Nice catch, so many edge cases. I'm going to start with some better error handling tomorrow probably, wouldn't have helped with this but for all the 1056546846 null checks

PR #352, opened 2024-09-23
feat/kubectx-menu

mosheavni:
Create a float view for `:Kubectx`
image

### DIFF
--- a/lua/kubectl/init.lua
+++ b/lua/kubectl/init.lua
@@ -1,3 +1,4 @@
+local ctx_view = require("kubectl.views.contexts")
 local informer = require("kubectl.actions.informer")
 local ns_view = require("kubectl.views.namespace")
 local state = require("kubectl.state")
@@ -105,12 +106,13 @@ function M.setup(options)
 
   vim.api.nvim_create_user_command("Kubectx", function(opts)
     if #opts.fargs == 0 then
-      return
+      ctx_view.View()
+    else
+      ctx_view.change_context(opts.fargs[1])
     end
-    completion.change_context(opts.fargs[1])
   end, {
     nargs = "*",
-    complete = completion.list_contexts,
+    complete = ctx_view.list_contexts,
   })
 end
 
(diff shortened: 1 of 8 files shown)


mosheavni:
ready for review @Ramilito

lua/kubectl/init.lua
```diff
@@ -105,12 +106,13 @@ function M.setup(options)
 
   vim.api.nvim_create_user_command("Kubectx", function(opts)
     if #opts.fargs == 0 then
-      return
+      ctx_view.View()
```
Ramilito, line 109: Nice!

lua/kubectl/mappings.lua
```diff
@@ -33,6 +33,7 @@ function M.register()
   M.map_if_plug_not_set("n", "<C-f>", "<Plug>(kubectl.filter_view)")
   M.map_if_plug_not_set("v", "<C-f>", "<Plug>(kubectl.filter_term)")
   M.map_if_plug_not_set("n", "<C-n>", "<Plug>(kubectl.namespace_view)")
+  M.map_if_plug_not_set("n", "<C-x>", "<Plug>(kubectl.contexts_view)")
```
Ramilito, line 36: Awesome mapping!

lua/kubectl/utils/viewsTable.lua
```diff
@@ -16,6 +16,8 @@ return {
   services = { "services", "service", "svc" },
   configmaps = { "configmaps", "configmap", "cm" },
   crds = { "customresourcedefinitions", "crds", "crd", "customresourcedefinitions.apiextensions.k8s.io" },
+  namespace = { "namespace", "namespaces", "ns" },
```
Ramilito, line 19: While I like the idea of adding namespaces here, that would remove it from the fallback views and we wouldn't be able to edit/delete them

lua/kubectl/views/contexts/init.lua
```diff
@@ -0,0 +1,77 @@
...
+  M.contexts = vim.split(contexts, "\n")
+  return M.contexts
+end
+
+--- Change context and restart proxy
+--- @param cmd string
+function M.change_context(cmd)
+  local results = commands.shell_command("kubectl", { "config", "use-context", cmd })
+
+  vim.notify(results, vim.log.levels.INFO)
```
Ramilito, line 67: This notify is making me have to press a key to continue, the results seem to be empty as well so I added a check for that and that works for me

lua/kubectl/utils/viewsTable.lua
mosheavni: You are right! i'll remove it

mosheavni, line 19:
```suggestion
```

Ramilito (approved):

PR #354, opened 2024-09-24
fix: table not filling space on larger window sizes

Ramilito:
We add extra padding to the last column, which is not wanted. This pr should address that so we fill the space properly.
 This only affects the table when the window is larger than the content.


mosheavni:
love that, works great for me except for a split second before the content is loaded where the columns are squished together

not super critical

Ramilito:
Thanks for testing it, didn't notice that, I guess that means on the first call, it's returning 0 as extra padding, will fix!

Ramilito:
this should now be correct and fill out the window to the edge, I could not replicate the error I see in your gif though, could try again? I'll merge this for now and if you get the same behaviour we can create a new PR

Ramilito:
Yeah that's a different padding that I'm getting an error removing but the extra padding (that is supposed to fill out the screen should be gone), the extra space you see should have been removed by the scrolloff though, are you on latest?

Ramilito:
This is how it's looking for me, if you ignore the 3 extra characters:
image

mosheavni:
> This is how it's looking for me, if you ignore the 3 extra characters: image

I do only have 3 extra spaces, less then before.

Ramilito:
Nice! We can probably get rid of those other 3 soon, it threw an error with the extmarks and those are a pain

Ramilito:
Managed to remove the extra 3 now as well!

PR #355, opened 2024-09-24
feat: add-fuzzy-completion

mosheavni:
discovered this:
https://github.com/swarn/fzy-lua
the guy just said we could copy the file, and it's amazing!!!
I love the result

next we should display a completion box with the suggestions, that could be even more amazing


Ramilito (approved):

mosheavni:
Yes sounds good

PR #363, opened 2024-09-25
feat/completion-pum

### LINKED ISSUE #357: feat: add completion box
see #355

mosheavni:
fixes #357 

WORK IN PROGRESS
image

### DIFF
--- a/lua/kubectl/actions/highlight.lua
+++ b/lua/kubectl/actions/highlight.lua
@@ -31,12 +31,13 @@ local highlights = {
   KubectlExperimental = { fg = "#CE9178" }, -- Brown
   KubectlNote = { fg = "#9CDCFE" }, -- Light Blue
   KubectlGray = { fg = "#666666" }, -- Dark Gray
+  KubectlPum = { bg = "#3e4451" },
   KubectlUnderline = { underline = true },
 }
 
 function M.setup()
   for group, attrs in pairs(highlights) do
-    local success, hl = pcall(api.nvim_get_hl_by_name, group, true)
+    local success, hl = pcall(api.nvim_get_hl, 0, group)
     if not success or not hl then
       api.nvim_set_hl(0, group, attrs)
     end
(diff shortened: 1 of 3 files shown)


mosheavni:
currently:


issues:

- [ ] long results are currently not shown (more than 50)
- [ ] pum stays visible for a few seconds after the selection is chosen
- [ ] vim.on_key is on every buffer. how can I limit to only the prompt buffer (alias/ns/ctx)?

@Ramilito  need ur help here 🙏🏼
you'd probably also want to move stuff around

Ramilito:
The gifs are looking soooo goood!
I'll try to get some time later tonight and look if I can help

mosheavni:
> The gifs are looking soooo goood!
> I'll try to get some time later tonight and look if I can help

Please work your magic here because I am really stuck and you seem to get the hang of all those UI components

Ramilito:
>  long results are currently not shown (more than 50)
 pum stays visible for a few seconds after the selection is chosen
 vim.on_key is on every buffer. how can I limit to only the prompt buffer (alias/ns/ctx)?

- Long results, I removed the limit you added and added a windows height restriction instead, so we would scroll for the rest.
- Not seeing the delay, is it still there?
- vim.on key wasn't called because the buffer wasn't really removed (just hidden), I set it to wipe on close and now it's getting called

Can you verify that I understood this correctly?

mosheavni:
yes, it looks great :)
I changed row to 1 to move the pum a line lower so it won't hide the text, and changed some stuff around the highlights. 
ok to merge? :)

lua/kubectl/actions/highlight.lua
```diff
@@ -31,12 +31,13 @@ local highlights = {
...
   KubectlNote = { fg = "#9CDCFE" }, -- Light Blue
   KubectlGray = { fg = "#666666" }, -- Dark Gray
+  KubectlPum = { bg = "#3e4451" },
   KubectlUnderline = { underline = true },
 }
 
 function M.setup()
   for group, attrs in pairs(highlights) do
-    local success, hl = pcall(api.nvim_get_hl_by_name, group, true)
+    local success, hl = pcall(api.nvim_get_hl, 0, group)
```
Ramilito, line 40: nice!

Ramilito (approved):

Ramilito:
It looks sooo cool!
Love your work here!

Issue #350, opened 2024-09-23
Bug: overview view crashing vim


Ramilito:
@mosheavni could you add some details?

I'm guessing it's probably too many curl commands at the same time or too many results

mosheavni:
first I'd truncate the nodes column, I have many nodes in my clusters and I'd want to see the top consuming and make all fit into one screen. Scrolling down is not as fun.

after a while, I'm just thrown out of vim, no errors, no nothing. next thing I see is my shell prompt, like this:
<img width="1308" alt="image" src="https://github.com/user-attachments/assets/70dfe932-bf34-4457-ae3d-a38187665d05">

Ramilito:
That's a lot of nodes 😅, displaying only top 10 now, also think I find the memory leak. Would be awesome if you could try it out?

mosheavni:
looks better, I'll see if it's crashing again.
<img width="425" alt="image" src="https://github.com/user-attachments/assets/d92b2d65-192b-4276-a2ce-54e09d936d56">
do you know why the column titles are going over the edge?

mosheavni:
also, just now:
<img width="1208" alt="image" src="https://github.com/user-attachments/assets/187bd5d6-3664-47f7-8550-8a58673498ac">

Ramilito:
Hmm not sure, will look into that.

Fixed the status nil classic error...

Ramilito:
@mosheavni the width should be similar to our normal table now as well.

mosheavni:
nice improvements!
events are overflowing since they are almost always long lines
i think events should be on their own line
<img width="991" alt="image" src="https://github.com/user-attachments/assets/6cbff0d7-ab6d-442f-8fb1-d9ef667f1531">

Ramilito:
Hmm this is a conundrum, the idea is that you are supposed to be able to choose how many columns you want on a row, so I can't guarantee that the events will be on it's own row. Added a config option for this.

How about I remove the object name and add a reasonable limit to the msg and a ... at the end if it exceeds the limit?

Ramilito:
@mosheavni Did a rework on how the grid layout is setup since the initial was to dynamically adjust based on a max_column per row setting, which made it harder to consider very long rows such as the events column.
The new implementation takes a layout instead in the get_headers section where you specify what columns go in what row, looks like this:
![image](https://github.com/user-attachments/assets/217d427d-3054-4cba-bb62-1e0d766b39c8)

[...]

PR #365, opened 2024-09-26
chore: remove notification feature

### LINKED ISSUE #248: Refactor/remove notifications
https://github.com/mosheavni/kubectl.nvim/blob/main/lua/kubectl/notification.lua

I tried to play with it, the view is really small, cannot see any relevant data in it, it disappears fast.
I think it's better to implement some logging mechanism, writing log lines to a file.
alot of times when I debug my side-branch, I have to add bunch of prints on `commands.shell_command_async()` to see what action was performed.
I'd essentialy enable debugging and then just read an external file for the plugin prints.
we can also add bunch of more debug prints after that is done.

Ramilito:
This feature is obsolete, the reason it was added was to show that the plugin was loading and working but now that we have the informer this isn't needed anymore.

Doesn't fix #248


mosheavni:
thank you! i didn't like it that much haha

Ramilito:
Yeah I agree!
I would love some feedback on the overview view, I suspect you aren't super happy with that either, if you could put the feedback here we can make it better: #350

PR #367, opened 2024-09-26
feat: configurable auto search

### LINKED ISSUE #366: toggle auto-filter
filter view previously didn't auto search when you choose from history,
I really liked being able to click `enter` on a filter then add/remove/change it and then apply it
Can we allow the user to choose whether to auto filter when clicking `enter` or just fill the prompt?
Also, when we just fill the prompt, the cursor is at the start of the line (we had this before, remember we had this issue before? @Ramilito)

So when the feature of auto-filter is turned off, I want the cursor to be at the end of the filter term.

Thanks buddy.

Ramilito:
Fixes #366 

Bonus:
- updated some configuration option in readme
- enabled object fresh by default
- fix set cursor to last item

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,15 +145,19 @@ vim.keymap.set("n", "<Plug>(kubectl.wrap)", "gw")
 {
   auto_refresh = {
     enabled = true,
-    interval = 3000, -- milliseconds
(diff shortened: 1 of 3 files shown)


mosheavni:
thanks!

Issue #368, opened 2024-09-26
configurable history


mosheavni:
I want to configure the max filter history to be 10
I want to configure the max alias history to be 5

thanks 🙏🏼

PR #370, opened 2024-09-26
feat: configurable max history

Ramilito:
Fixes: #368 
Bonus:
- Fix delay of closing pum window
- Fix hint mappings not displaying (still globals are not visible)


mosheavni:
nice!

PR #374, opened 2024-09-26
feat/use-data-for-resource

mosheavni:
on k8s_ ftplugins on some keymaps, we get the user selection and then do `kubectl get resource name -n namespace -ojson`

but why?
we already have all the resource data!
`find_resource` function to the rescue

this makes the stuff changed blazingly fast.
change image, scale deployment, browse ingress
we don't have to wait to fetch the resource again, we just search for it.

it looks like a big PR with many changes but in reality I just went outside of the resource builder `fetchAsync()` that's why I had to indent everything back and it looks like many changes.

try it @Ramilito  i think you'd like it

### DIFF
--- a/ftplugin/k8s_daemonsets.lua
+++ b/ftplugin/k8s_daemonsets.lua
@@ -1,10 +1,11 @@
 local api = vim.api
-local ResourceBuilder = require("kubectl.resourcebuilder")
 local buffers = require("kubectl.actions.buffers")
 local commands = require("kubectl.actions.commands")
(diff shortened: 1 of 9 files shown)


Ramilito:
It's absolutely a good design choice, saves us api calls and is basically net even with lines of codes.
I created a find:single that is supposed to return an item from a dictionary but having a kubernetes specialized find is also pretty awesome.
Lets merge it!

Ramilito (approved):

PR #375, opened 2024-09-26
Feat/label selector

### LINKED ISSUE #242: feat: apply label filters
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Sometimes I grab a label from somewhere and want to see pods or ingresses or whatever that relates to that label
Maybe some small float/input box for applying a labelSelector to the current view?
some of the logic is already here:
https://github.com/Ramilito/kubectl.nvim/pull/225/files#diff-8f3fd4a79f3dccfbc4225f7a209dee6abd01efdbb3dcacd6ce66ac372cf5e71bR170

**Describe alternatives you've considered**

**Additional context**

mosheavni:
WORK IN PROGRESS
fixes #242


mosheavni:
it's working when the definition.cmd is curl
it's not working when opening the plugin on a fallback resource and the cmd is kubectl even though I add -l app=my-app to the definition.args

any idea why not? @Ramilito

mosheavni:
Also when the cached_resources are loaded, and the fallback is now `curl`, the same filter that's working for other resources doesn't work on fallback

mosheavni:
it's because we define `definition.cmd` and `definition.url` every time `View()` is called. working on it.

Ramilito:
I had a bunch of stuff to do but now I'm back so let me know if you need me, was going to try to finish #354 otherwise!

mosheavni:
Ready for review @Ramilito 

it's working.
I'd like to add highlight in the future so key=value - key would be highlighted green. do you think this is easy to highlight when a user exits insert mode?

mosheavni:
also I'd like to show the user that he is seeing resources with labels somewhere, maybe in the context bar?

Ramilito:
I'm testing it out right now and It's looking really awesome!




> I'd like to add highlight in the future so key=value - key would be highlighted green. do you think this is easy to highlight when a user exits insert mode?

The super easy way is to set a filetype that you feel is correct, that might not work here if you can't stick to yaml or json.
Otherwise, I would need to look into it but not sure what you mean with exits insert mode?


> also I'd like to show the user that he is seeing resources with labels somewhere, maybe in the context bar?

Some are saving space by hiding the context and hints bar and it's too much information to put it in the divider like we use the filter for

mosheavni:
ok we can do it in a different PR
can we merge?

lua/kubectl/utils/tables.lua
```diff
@@ -386,12 +386,12 @@ end
...
   if data.items then
     return vim.iter(data.items):find(function(row)
-      return row.metadata.name == name and row.metadata.namespace == namespace
+      return row.metadata.name == name and (namespace and row.metadata.namespace == namespace or true)
     end)
   end
   if data.rows then
     return vim.iter(data.rows):find(function(row)
-      return row.object.metadata.name == name and row.object.metadata.namespace == namespace
+      return row.object.metadata.name == name and (namespace and row.object.metadata.namespace == namespace or true)
```
Ramilito, line 394: Am I thinking wrong but won't this always return true:
`(namespace and row.object.metadata.namespace == namespace or true)`

lua/kubectl/views/filter/init.lua
```diff
@@ -31,7 +34,77 @@ local function save_history(input)
   state.filter_history = result
 end
 
+function M.filter_label()
+  local state = require("kubectl.state")
+  local instance = vim.deepcopy(state.instance)
```
Ramilito, line 39: Do we need to deepcopy here? this object can be really big!

lua/kubectl/views/filter/init.lua
```diff
@@ -31,7 +34,77 @@ local function save_history(input)
...
+  local view, definition = views.view_and_definition(instance.resource)
+  local name, ns = view.getCurrentSelection()
+  if not name then
+    return
+  end
+
+  local resource = tables.find_resource(require("kubectl.state").instance.data, name, ns)
```
Ramilito, line 46: Don't we already have the instance from above?

Ramilito:
> ok we can do it in a different PR can we merge?

Left a few comments that I was curious about, would love it if  you could resolve those so we can merge it!

lua/kubectl/utils/tables.lua
mosheavni: I tested with this file:

```lua
---@diagnostic disable: undefined-global
local M = {
	completions = {},
}

-- Reload the file when it changes on disk
local group = vim.api.nvim_create_augroup("ReloadModule", { clear = true })
vim.api.nvim_create_autocmd("BufWritePost", {
	buffer = 0,
	group = group,
	callback = function()
		vim.cmd("luafile %")
		vim.notify("Module reloaded")
	end,
})

vim.keymap.set("n", "bla", function()
	-- Define the user command

	local data = {
		rows = {
			{
				object = {
					metadata = {
						name = "test",
						namespace = "default",
					},
				},
			},
			{
				object = {
					metadata = {
						name = "no-ns",
					},
				},
			},
		},
	}
	local name = "test"
	local namespace = "default"
	-- local name = "no-ns"
	local hiri = vim.iter(data.rows):find(function(row)
		return row.object.metadata.name == name and (namespace and row.object.metadata.namespace == namespace or true)
	end)
	vim.notify(vim.inspect(hiri))
end)
```

- save the file
- open it
- run `:so%` to source the file
- click `bla` to trigger the function - now you see the object with name and namespace
- comment out name and namespace
- comment in no-ns
- save (it will reload the module)
- click `bla` - now you see the object without the namespace

Maybe it's not the smartest(?) but it's working

lua/kubectl/views/filter/init.lua
mosheavni: it's working without deepcopy. I removed it and added highlights

lua/kubectl/views/filter/init.lua
mosheavni: fixed

Ramilito (approved):

Ramilito:
This feature might be useful later  together with kubectl tree feature

PR #376, opened 2024-09-27
feat/remove-strings-trim

mosheavni:
use `vim.trim` instead of the `string_utils.trim`


Ramilito (approved):

Ramilito:
So many hidden vim utilities, nice find

PR #395, opened 2024-09-28
Feat/keymap view names

mosheavni:
Change the name for views keymaps, also add every view and define them dynamically
I tested 2 keymaps that don't have any default and it's working.


Ramilito (approved):

Ramilito:
This is a great way to make it easier to customize! With less code even!

PR #403, opened 2024-09-29
feat/negative-and-multi-filter

mosheavni:
since I would not expect to find a comma `,` or an exclamation sign `!` I used them as the tokens
negative search with `!`
multiple searches concatenated by a comma

also look at the gif @Ramilito 
you can see that the hints are not populated in the first time I open filter float, only on the second time we can see the hints.

I guess this issue would not leave us alone.

### DIFF
--- a/lua/kubectl/utils/find.lua
+++ b/lua/kubectl/utils/find.lua
@@ -76,23 +76,42 @@ function M.filter(tbl, predicate)
   return result
 end
 
---- Filter lines in an array based on a pattern starting from a given index
+--- Filter lines in an array based on multiple patterns starting from a given index
 ---@param array table[]
----@param pattern string
+---@param patterns string
 ---@param startAt number
 ---@return table[]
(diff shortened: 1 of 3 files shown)


lua/kubectl/utils/find.lua
```diff
@@ -76,23 +76,41 @@ function M.filter(tbl, predicate)
...
     local line = array[index]
-    if M.is_in_table(line, pattern) then
+    local all_match = true
+
+    for _, pattern in ipairs(pattern_list) do
+      local is_negative = pattern:sub(1, 1) == "!"
+      local actual_pattern = is_negative and pattern:sub(2) or pattern
+      local match = M.is_in_table(line, actual_pattern)
+
+      if (is_negative and match) or (not is_negative and not match) then
```
Ramilito, line 108: This is a great idea 👍

Ramilito (approved):

PR #415, opened 2024-09-30
chore/pod-container-count

mosheavni:
we calculate the number of containers from the `row.status` but if the pod was evicted, we won't get any `containerStatuses` which makes the pod's READY column to be `0/0` thus being colored blue which is good when in fact it's not.

Better to calculate the number of containers from `row.spec`


Ramilito:
Ah clever!

Ramilito (approved):

Issue #407, opened 2024-09-30
Bug: hints not populated on first load


Ramilito:
Mentioned here: #403

PR #416, opened 2024-09-30
fix: hints not visible on first load for floats

Ramilito:
Fixes: #407 

The diff is crazy, I just wrapped it with vim.schedule...


mosheavni:
great!

PR #417, opened 2024-09-30
chore/use-plug-on-refresh

mosheavni:
Use `<Plug>(kubectl.refresh)`

TBH I'm not sure if we need it, but just in case.


Ramilito:
I like it!

Ramilito (approved):

PR #418, opened 2024-09-30
feat/add-close-keymaps

mosheavni:
add mappings to close a float buffer
<kbd>ctrl-c</kbd> on insert mode
<kbd>q</kbd> and <kbd>ESC</kbd> on normal mode

I really feel those are defaults and we can avoid letting it being changed.

### DIFF
--- a/lua/kubectl/actions/buffers.lua
+++ b/lua/kubectl/actions/buffers.lua
@@ -3,6 +3,22 @@ local state = require("kubectl.state")
 local api = vim.api
 local M = {}
 
+local function close_keymaps(buf, change_modified)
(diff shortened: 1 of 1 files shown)


Ramilito:
Did another pr as a  suggestion: #419 , what do you think? which should we use?

mosheavni:
> Did another pr as a suggestion: #419 , what do you think? which should we use?

yes yours is better! just make sure the `<C-c>` mapping is this:

```lua
      vim.api.nvim_buf_set_keymap(0, "i", "<C-c>", "<Esc><Plug>(kubectl.quit)", opts)

```

PR #429, opened 2024-10-04
feat: fetch api resources in a efficient way

Ramilito:
Implement fetchAllAsync to resourcebuilder and use that for getting all the api-resources at the same time.

### DIFF
--- a/lua/kubectl/resourcebuilder.lua
+++ b/lua/kubectl/resourcebuilder.lua
@@ -100,6 +100,17 @@ function ResourceBuilder:fetch()
   return self
 end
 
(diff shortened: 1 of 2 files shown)


mosheavni:
nice! did you check the difference of the execution time? is it significantly better?

Ramilito:
Only the first two calls would have been impacted, since we did a fetch for the normal apis (`/apinv1`) then when that was done we would get the second group (`/apis`). getting the other stuff was done in a similar way since the fetchAllAsync basically just loops over a list and runs fetch_async.
This now does the first step at the same time.

PR #430, opened 2024-10-04
feat: lineage

### LINKED ISSUE #130: feat: hierchial view
**Describe the solution you'd like**

I way to visualize the whole app hierarchically.

**Describe alternatives you've considered**

Either we do this in this plugin or use an external tool such as kubectl lineage

Ramilito:
Fixes #130
Basic functionality seems to be working, need to dig deeper into the performance impact at startup, after the initial load, the action is very fast.
Does not block startup and doesn't delay the aliases so might be okay if we do this in the background.

- [x] Collect all owner information in a cache
- [x] Improve curl command to use --parrallel together with a range of urls to so we can reuse the same tcp connection and only use one shell cmd -- not possible since the output is jumbled, reverted to running multiple curl instead
- [x] Create a graph of relationship 
- [x] Present results 
- [ ] how to refresh cache? informer? timer?
    - [x] manually/on demand
    - [ ] automatic? On some event maybe? don't want to run it too often if the feature isn't being used. @mosheavni what do you think?
- [x] use label selectors for more relationships
- [x] add event relationships
- [x] make work with fallbackview
- [x] batch api calls
- [x] Memory impact? -- Acceptable, ~4k pods  300mb
- [x] Add go to resource feature
- [x] color output
- [x] add folding
- [x] Reload cache on context switch
- [x] Add config options to turn this on/off and a limit of sort
- [x] Selector not working unless parent has the same labels
- [x] Ingress not showing up
- [ ] Go to should narrow the selection down.
- [x] Sort output to avoid jumping text

lineage


mosheavni:
nice :) although right now i'm getting this:
image

if you're still in progress and that's expected, then disregard :)

mosheavni:
lineage shows `Cache still loading...` and the alias menu doesn't have all of the resources.
no matter how long I wait, I still don't have all of the cache

Ramilito:
The alias content should be done before we start with the lineage stuff, could you check if it's missing in main as well? I might have messed up in the refactor

mosheavni:
actually now it's working nicely
image

do you think we could add a mapping to go to the resource?

mosheavni:
also maybe color resource types in different colors to make it more distinguishable?

Ramilito:
I liked all the ideas! Added them to the todo list

mosheavni:
results are much better now! wow!

mosheavni:
`kubectl lineage` for service shows endpointSlice and ingress.
your lineage shows just the endpointSlice.
Why did it miss the ingress?

Ramilito:
The ownership results seem to be working well, I think that's what you are seeing, the services need the labels selection that I removed when rewriting it, Il add it back soon and I think ingress has a third lookup not sure yet, Il check when the label selection is done if they show up

mosheavni:
I think folding could be implemented easily by setting `foldmethod=indent` since you're already indenting the table we could benefit from this build-in capability of folding by indent levels

Ramilito:
> I think folding could be implemented easily by setting `foldmethod=indent` since you're already indenting the table we could benefit from this build-in capability of folding by indent levels

Added!

mosheavni:
indeed I see all resources now. what a great work you did here

Ramilito:
> indeed I see all resources now. what a great work you did here

There is still a bug in the label selector, ingress is still not done but it's getting close now!

lua/kubectl/actions/highlight.lua
```diff
@@ -32,12 +33,24 @@ local highlights = {
...
   KubectlGray = { fg = "#666666" }, -- Dark Gray
+  KubectlWhite = { fg = "#FFFFFF" }, -- White
   KubectlPselect = { bg = "#3e4451" }, -- Grey Blue
   KubectlPmatch = { link = "KubectlWarning" },
   KubectlUnderline = { underline = true },
 }
 
+local function add_bold_variant()
+  for group, attrs in pairs(highlights) do
+    highlights[group .. "Bold"] = vim.tbl_extend("force", attrs, { bold = true })
```
Ramilito, line 44: I think this is running multiple times, need to check this.

Ramilito:
Rewrote this implementation (for the 45645645 time) and think this is the last one, the generic lookup with labels took me very far but in the end it wasn't reliable nor enough to cover all cases.
The latest way is generic in some parts but all special relationships need to be specified. I think I solved it in a scalable way using the relationship.lua table mapping, going to expand on that list and see where that takes me.

mosheavni:
Wow I gotta take a deep dive here
Seems like holding a relationship map is the best way since resources are related to other resources in a bunch of different ways. I'll definitely add some more relation. Mind if I test a bit before merging?

lua/kubectl/utils/timeme.lua
```diff
@@ -7,8 +7,8 @@ end
 
 M.stop = function()
   local elapsed_time = vim.fn.reltimefloat(vim.fn.reltime(M.startTime))
-  local log = require("kubectl.log")
-  log.fmt_info("Execution time: %.6f seconds", elapsed_time)
+  -- local log = require("kubectl.log")
```
Ramilito, line 10: Revert this!

Ramilito:
> Wow I gotta take a deep dive here
> Seems like holding a relationship map is the best way since resources are related to other resources in a bunch of different ways. I'll definitely add some more relations for custom resources that are heavily used. Mind if I test a bit before merging?

Yeah exactly, it's crazy how much relationships I take for granted that, the relationship map is at least architecturally easy to manage and scale, a pain to refactor and I did not plan to add 1k lines of code for this 😓

I would love it if you could test and try to expand on it before merging, I'm also setting lineage to be off by default and maybe setting a beta tag on this feature when we merge it.

PR #439, opened 2024-10-08
chore: misc doc updates and remove unused `exact` param from is_in_table

paulbrittain:
- Misc function doc updates & fixes as I read through code
- Code change:`lua/kubectl/utils/find.lua` is_in_table() removed unused `exact` param
- Code change: Set `stdin` param of `vim.system` call to nil instead of false - suggested by LSP (May be wrong?)

### DIFF
--- a/lua/kubectl/actions/kube.lua
+++ b/lua/kubectl/actions/kube.lua
@@ -40,7 +40,7 @@ function M.start_kubectl_proxy(callback)
   local handle = vim.system(command.args, {
     clear_env = true,
     env = command.env,
-    stdin = false,
+    stdin = nil,
     stderr = function(_, data)
       vim.schedule(function()
         if data then
--- a/lua/kubectl/actions/kube.lua
+++ b/lua/kubectl/actions/kube.lua
@@ -40,7 +40,6 @@ function M.start_kubectl_proxy(callback)
   local handle = vim.system(command.args, {
     clear_env = true,
     env = command.env,
-    stdin = nil,
     stderr = function(_, data)
       vim.schedule(function()
         if data then
--- a/lua/kubectl/actions/kube.lua
+++ b/lua/kubectl/actions/kube.lua
@@ -40,6 +40,7 @@ function M.start_kubectl_proxy(callback)
   local handle = vim.system(command.args, {
     clear_env = true,
     env = command.env,
+    stdin = false,
     stderr = function(_, data)
       vim.schedule(function()
         if data then
--- a/lua/kubectl/actions/layout.lua
+++ b/lua/kubectl/actions/layout.lua
@@ -130,7 +130,11 @@ function M.get_editor_dimensions()
   return width, height
 end
 
-function M.win_size_fit_content(buf_nr, offset)
+--- Fits content to window size
+--- @param buf_nr integer: The buffer number.
+--- @param height_offset integer: The height offset.
+--- @return vim.api.keyset.win_config
+function M.win_size_fit_content(buf_nr, height_offset)
   local win_id = vim.api.nvim_get_current_win()
   local win_config = vim.api.nvim_win_get_config(win_id)
 
@@ -146,10 +150,10 @@ function M.win_size_fit_content(buf_nr, offset)
     end
   end
 
-  win_config.height = rows + offset
+  win_config.height = rows + height_offset
   win_config.width = max_columns
 
-  api.nvim_set_option_value("scrolloff", rows + offset, { win = win_id })
+  api.nvim_set_option_value("scrolloff", rows + height_offset, { win = win_id })
   vim.api.nvim_win_set_config(win_id, win_config)
   return win_config
 end
--- a/lua/kubectl/actions/layout.lua
+++ b/lua/kubectl/actions/layout.lua
@@ -133,7 +133,7 @@ end
 --- Fits content to window size
 --- @param buf_nr integer: The buffer number.
 --- @param height_offset integer: The height offset.
---- @return vim.api.keyset.win_config
+--- @return { height: number, width: number }
 function M.win_size_fit_content(buf_nr, height_offset)
   local win_id = vim.api.nvim_get_current_win()
   local win_config = vim.api.nvim_win_get_config(win_id)
--- a/lua/kubectl/utils/find.lua
+++ b/lua/kubectl/utils/find.lua
@@ -34,7 +34,7 @@ function M.is_in_table(tbl, str, exact)
   local lowered_str = str:lower()
   for _, value in pairs(tbl) do
     if type(value) == "table" then
-      if M.is_in_table(value, str, exact) then
+      if M.is_in_table(value, str, false) then
         return true
       end
     end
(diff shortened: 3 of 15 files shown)


lua/kubectl/actions/layout.lua
```diff
@@ -146,10 +150,10 @@ function M.win_size_fit_content(buf_nr, offset)
     end
   end
 
-  win_config.height = rows + offset
+  win_config.height = rows + height_offset
```
Ramilito, line 153: nice!

lua/kubectl/utils/find.lua
```diff
@@ -25,7 +25,7 @@ end
 ---@param tbl table
 ---@param str string
 ---@return boolean
-function M.is_in_table(tbl, str, exact)
+function M.is_in_table(tbl, str)
```
Ramilito, line 28: Are we not using exact here?: https://github.com/Ramilito/kubectl.nvim/blob/e0bfb6eda2248e5f4ef83cf35cca44e95645e253/lua/kubectl/views/init.lua#L293

lua/kubectl/utils/loop.lua
```diff
@@ -7,7 +7,7 @@ local active_sessions = {}
 --- Start a loop for a specific buffer
 ---@param buf number
 ---@param callback fun(is_cancelled: fun(): boolean)
----@param opts? table
+---@param opts? { interval: number }: The arguments for the loop.
```
Ramilito, line 10: nice!

lua/kubectl/utils/tables.lua
```diff
@@ -3,6 +3,7 @@ local hl = require("kubectl.actions.highlight")
 local state = require("kubectl.state")
 local time = require("kubectl.utils.time")
 local M = {}
+local context_rows = {}
```
Ramilito, line 6: I don't think this is supposed to be a global (for the module) variable?

lua/kubectl/utils/tables.lua
```diff
@@ -352,7 +361,6 @@ function M.generateHeader(headers, include_defaults, include_context, divider)
   end
 
   -- Add context rows
-  local context_rows = {}
```
Ramilito, line 355: I think this is the one I mentioned above ☝️

lua/kubectl/actions/kube.lua
```diff
@@ -40,7 +40,7 @@ function M.start_kubectl_proxy(callback)
   local handle = vim.system(command.args, {
     clear_env = true,
     env = command.env,
-    stdin = false,
+    stdin = nil,
```
Ramilito, line 43: It should be false according to the docs, not sure why lsp is saying something else?

image

mosheavni:
Nice work here thank you. Just fix @Ramilito comments :)

lua/kubectl/actions/kube.lua
paulbrittain: My interpretation was that it's evaluated to false if nil. Maybe it's clearer to simply not pass `stdin` when not using it?

lua reference:
`--- @field stdin? string|string[]|true`

lua/kubectl/utils/tables.lua
paulbrittain: It was being accessed by `addContextRows` as a global variable, but it's been refactored in #440 now anyway 👍

lua/kubectl/utils/find.lua
paulbrittain: True. My apologies. Not sure what happened to tooling last night, will look into why and fix.

Reverted.

I did notice that the `exact` param is not being passed to the recursive call, though, resulting in inconsistent `exact` matching with nested tables. What do you think?

lua/kubectl/actions/kube.lua
mosheavni: better be explicit than implicit no? @paulbrittain

lua/kubectl/utils/find.lua
mosheavni: we had problems with it:
https://github.com/Ramilito/kubectl.nvim/issues/308

mosheavni:
also take a look at the type check that is failing:
https://github.com/Ramilito/kubectl.nvim/actions/runs/11249290844/job/31276754598?pr=439

lua/kubectl/utils/find.lua
paulbrittain: changes to explicitly pass false for recursive calls

lua/kubectl/actions/kube.lua
paulbrittain: Fair enough, then it's a Lua typing issue that causes this ugly assignment mismatch diagnostic because it's not typed `|boolean`. I'll just revert to false  :)

image

paulbrittain:
> also take a look at the type check that is failing: https://github.com/Ramilito/kubectl.nvim/actions/runs/11249290844/job/31276754598?pr=439

Will do. How do you run the type check locally? :)

lua/kubectl/actions/kube.lua
Ramilito: Yeah your interpretation was probably correct @paulbrittain , I must confess that it's a weird thing to support only type true and not allow false when the default value later on is false....if you still have this error, I don't mind using nil here since it's basically the same as false.

Ramilito:
> > also take a look at the type check that is failing: https://github.com/Ramilito/kubectl.nvim/actions/runs/11249290844/job/31276754598?pr=439
> 
> Will do. How do you run the type check locally? :)

Like this:
```zsh
luarocks install luacheck
luacheck path/to/your/project
```

lua/kubectl/actions/kube.lua
mosheavni: can't we do a comment casting? something like `---@as boolean`?

Ramilito: I'm fine with both ways! we should aim not to have any warnings if possible, they can easily get out of hand and then no one knows what's important or a false alarm

Ramilito: I agree on passing nil, that should also satisfy @mosheavni point on being explicit!

paulbrittain:
@Ramilito Hey Ramilito, yes the local linter doesn't return the same lint errors as the pipeline. Can the linting workflow be enabled so it runs on commits so that I can iterate on resolving the errors it returns commit-by-commit?

Is there a cost incurred per run?

Ramilito:
> @Ramilito Hey Ramilito, yes the local linter doesn't return the same lint errors as the pipeline. Can the linting workflow be enabled so it runs on commits so that I can iterate on resolving the errors it returns commit-by-commit?
> 
> Is there a cost incurred per run?

No costs yet so all good, there is a safeguard for all new contributors, probably to avoid denial of service attacks so I had to press a button to allow you to run workflows! Button pressed!

lua/kubectl/actions/layout.lua
```diff
@@ -130,7 +130,11 @@ function M.get_editor_dimensions()
   return width, height
 end
 
-function M.win_size_fit_content(buf_nr, offset)
+--- Fits content to window size
+--- @param buf_nr integer: The buffer number.
+--- @param height_offset integer: The height offset.
+--- @return vim.api.keyset.win_config
```
mosheavni, line 136: what's this return?
i think it should be:
```
--- @return { height: number, width: number }
```

mosheavni (changes requested):

paulbrittain:
Ready for another workflow run :^)

Ramilito (approved):

Ramilito:
LGTM! Awesome work @paulbrittain, if you are happy with it @mosheavni we can merge it!

mosheavni (approved):

PR #440, opened 2024-10-08
chore: refactor versions to use max width

Ramilito:
Reuse current width calculator.
Prepare for more columns


Ramilito:
@mosheavni  found a simpler way to do the width thing, should be same functionality as the one you did though!

PR #441, opened 2024-10-09
Feat/toggle headers

mosheavni:
Add keymaps to toggle the headers ad-hoc.
Sometimes I just want to hide the headers instead of going to the config and disabling and then re-enabling.
Add some simple mappings to toggle the headers with the option/alt key


Ramilito:
I like it but how about just one keymap that toggles the whole header?

mosheavni:
Done. do you like it better?

Ramilito (approved):

Ramilito:
I love it!

PR #443, opened 2024-10-10
chore/better-lineage-fold

mosheavni:
related: #430


Ramilito:
Awesome! Tried something similar but couldn't get it to work

PR #446, opened 2024-10-13
Feat/log options configurable

mosheavni:
make log options configurable.


lua/kubectl/views/containers/init.lua
```diff
@@ -25,6 +26,14 @@ function M.exec(pod, ns)
 end
 
 function M.logs(pod, ns, reload)
+  local since_last_char = string.sub(M.log_since, -1)
```
Ramilito, line 29: This is to convert it to number because container logs only works with numbers right? But don't we need to multiply it then? Sorry if I'm just missing it, I'm on my phone while my kid is trying to draw stuff on my face

Ramilito (approved):

lua/kubectl/views/containers/init.lua
mosheavni: yeah it's just below it. lines 30-36

Ramilito (approved):

PR #447, opened 2024-10-15
fix: add check for str in capitalize func

paulbrittain:
**Description**

Fixes error when an empty string is passed to string util `capitalize` function.

**Reproduction**

This will occur when:

1. Go to a container view of a pod
2. `C-F` to filter
3. Press enter without entering in filter text

`lua/kubectl/mappings.lua`
```
pcall(view[string_utils.capitalize(resource[2])], resource[3] or "", resource[4] or "")
```

This call to `capitalize` will pass `resource[3] or ""`.  `resource[3]` will not be present so "" is passed as an argument.

**Error**

```
Error  11:50:54 msg_show.emsg E5108: Error executing lua: ...ttain/Personal/kubectl.nvim/lua/kubectl/utils/string.lua:5: attempt to index local 'str' (a nil value)
stack traceback:
	...ttain/Personal/kubectl.nvim/lua/kubectl/utils/string.lua:5: in function 'capitalize'
	...lbrittain/Personal/kubectl.nvim/lua/kubectl/mappings.lua:166: in function <...lbrittain/Personal/kubectl.nvim/lua/kubectl/mappings.lua:139>
```

PS:
1. I'd have added a test as a habit, but there are none. I checked telescope and fugitive and there aren't unit tests really. I guess this is normal for plugins? I did some googling around and found that [plenary.nvim](https://github.com/nvim-lua/plenary.nvim/blob/master/TESTS_README.md) has a testing framework.
2. Filtering containers isn't working for me, but I don't know why yet. For example, when I have a list of containers and one has the name "foo", when I enter "foo" into the filter I expect all other containers to be filtered out. But all containers are still visible.

### DIFF
--- a/lua/kubectl/utils/string.lua
+++ b/lua/kubectl/utils/string.lua
@@ -2,7 +2,11 @@ local M = {}
 
 ---@param str string
 function M.capitalize(str)
-  return str:sub(1, 1):upper() .. str:sub(2)
+    if str and #str > 0 then
+        return str:sub(1, 1):upper() .. str:sub(2)
+    else
+        return str
+    end
 end
 
 ---@param buf number The buffer number.


mosheavni:
LGTM

Ramilito:
LGTM!
Awesome PR, great details and very clear 🙏 

> 1. I'd have added a test as a habit, but there are none. I checked telescope and fugitive and there aren't unit tests really. I guess this is normal for plugins? I did some googling around and found that [plenary.nvim](https://github.com/nvim-lua/plenary.nvim/blob/master/TESTS_README.md) has a testing framework.


We don't want to add plenary since it's kind of unnecessary, we wouldn't use anything else it has either.
 
> 2. Filtering containers isn't working for me, but I don't know why yet. For example, when I have a list of containers and one has the name "foo", when I enter "foo" into the filter I expect all other containers to be filtered out. But all containers are still visible.

You are correct, it's been an active choice to not filter on floats, see the false here: https://github.com/Ramilito/kubectl.nvim/blob/69109da65c868bde689d8af845a16f0282802aa7/lua/kubectl/resourcebuilder.lua#L312 since the information is usually not that large, but maybe you have that use case?

Ramilito (approved):

Ramilito:
For the testing, I've been thinking [busted](https://github.com/lunarmodules/busted) might fit good, you are more than welcome to evaluate it!

paulbrittain:
> but maybe you have that use case?

I don't actually, but the hint being visible in the float suggests that it's a feature. We should not show hints when they're not applicable to the buffer. What do you think @Ramilito ? 😃 

>  you are more than welcome to evaluate it!

I can do that 👍 What criteria are we interested in evaluating? Or are you looking for an example usage in a branch of this plugin?

Ramilito:
> I don't actually, but the hint being visible in the float suggests that it's a feature. We should not show hints when they're not applicable to the buffer. What do you think @Ramilito ? 😃

I totally agree, removed global hints on all now in #448 

> I can do that 👍 What criteria are we interested in evaluating? Or are you looking for an example usage in a branch of this plugin?

Some notes I would want answered before committing to busted would be:
- Does it affect the user in any way? I am guessing no but you never know 
- Can it be run in github actions?
- Do we need to mock data for it to run if there is no k8s context?
- What the limit? can it handle complex ui states? can it validate nvim stuff or is it just for pure lua functions?

I'm not sure these are dealbreakers yet but we should make a conscious decision about it

Issue #435, opened 2024-10-07
feat: Select multiple resources to act on


Keyruu:
### Discussed in https://github.com/Ramilito/kubectl.nvim/discussions/433

I would love to select multiple pods and then delete all of them. Is that possible?</div>

mosheavni:
nice feature idea

PR #458, opened 2024-10-18
feat: multiselect actions

Ramilito:
Fixes: #435 

- [x] Act on visually selected lines
- [x] Make getCurrentSelection able to handle both visual and normal
- [ ] Refactor kubectl.kill to reuse the logic between normal and visual binding
- [ ] Investigate other actions we want to do on multirow
- [ ] Make it work for all other resources

### DIFF
--- a/ftplugin/k8s_pods.lua
+++ b/ftplugin/k8s_pods.lua
@@ -43,6 +43,37 @@ local function set_keymaps(bufnr)
     end,
   })
 
+  api.nvim_buf_set_keymap(bufnr, "v", "<Plug>(kubectl.kill)", "", {
+    noremap = true,
+    silent = true,
+    desc = "delete pod",
+    callback = function()
+      local rows = pod_view.getCurrentSelection()
+      if not rows then
+        return
+      end
+
+      for _, row in ipairs(rows) do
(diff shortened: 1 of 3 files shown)


mosheavni:
I wanna give an idea
Instead of visually select and the visual selection could change on every refresh and select different resources, how about "marking" the selected pods maybe with <tab> and adding a mark (just like the port forward symbol) and then if the view refreshes and the order changes, or even when you sort differently, the selection would stay the same!
Nvim-tree is doing something like that with a concept called "bookmarks"
I know this is different that what you started writing but it's far more reliable

mosheavni:
Visual selection is static, the pods can change their location on the table but the visual selection would stay the same since in vim the selection is static, but our views are not static by definition

Ramilito:
Thank you for the feedback and it's ok if already started.
I like your idea, I'll try it out

mosheavni:
If adding a mark to a resource and easily retrieving it after, it could be the best way. I wanna help. I'll try a bit to mess and we can compare results:)

Ramilito:
There is actually something I used to get the sorting to work properly that is using marks in a similar way, check out the utils/marks.lua file

Issue #463, opened 2024-10-20
feat: implement `kubectl debug` in containers view


mosheavni:
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_debug/

we can implement it on the containers view

Ramilito:
Ooo that's a great idea, would be really useful, I find typing it in kubectl is always a hassle

PR #471, opened 2024-10-22
feat: add debug command to container

Ramilito:
Fixes: #463 
@mosheavni This is working pretty smooth, any suggestions for improvements?

### DIFF
--- a/ftplugin/k8s_containers.lua
+++ b/ftplugin/k8s_containers.lua
@@ -6,6 +6,20 @@ local tables = require("kubectl.utils.tables")
 
 --- Set key mappings for the buffer
 local function set_keymaps(bufnr)
+  api.nvim_buf_set_keymap(bufnr, "n", "<Plug>(kubectl.debug)", "", {
+    noremap = true,
+    silent = true,
+    desc = "Debug",
+    callback = function()
+      local container_name = tables.getCurrentSelection(unpack({ 1 }))
+      if container_name then
+        container_view.selectContainer(container_name)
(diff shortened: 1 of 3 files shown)


mosheavni:
I always want to be able to choose the image, busybox is very limited

Ramilito:
Added prompt for image name but while testing it I found out that the ephemeral containers stick around even after the session terminates and it seems to be intentional for audit purposes according to discussions in Kubernetes github.

How do you work around that? Right now you can only debug once per container basically and then need to restart the pod

mosheavni:
how do you see it? can you send a screenshot?
from the docs, it's not possible to do anything with the ephemeral container once created so not something that should bother us.
kubernetes users should take that into consideration before creating them.

I'd make the image selection as a float like the drain node float, just to keep consistency, but all in all looks great.

Ramilito:
> how do you see it? can you send a screenshot? from the docs, it's not possible to do anything with the ephemeral container once created so not something that should bother us. kubernetes users should take that into consideration before creating them.
> 
> I'd make the image selection as a float like the drain node float, just to keep consistency, but all in all looks great.

This is the pod after you have created and exited a debug session:

image

And this is the second time you try to connect:
image


Maybe we just need to add the option to set the debug container name, that would at least remove the need to restart the pod for a new debug session

mosheavni:
maybe we can add the ephemeral container to the container float that way we could click `<CR>` and easily exec back into it?

Ramilito:
Hmm that's an interesting idea

Issue #462, opened 2024-10-20
feat: drain node with flags


mosheavni:
most commonly when I drain a node I use the flags `--ignore-daemonsets` and `--delete-emptydir-data`
sometimes also `--force` and `--grace-period`

we could use something like k9s offers:
<img width="587" alt="image" src="https://github.com/user-attachments/assets/addc72fa-4de5-44e4-bf78-60a34ff7c384">

PR #473, opened 2024-10-22
feat: drain node with flags

Ramilito:
Fixes: #462


Ramilito:
This is working, will move some code around before merging

mosheavni:
ohhh so beatiful!
any idea why the highlights mess up when I'm changing fields?

Ramilito:
Love it, thank you for the suggestion, totally forgot about that approach and got tunnel vision using extmarks :D.

PR #480, opened 2024-10-24
feat: generic action view

Ramilito:
- [x] convert existing functionality to be used generically
- [x] make node drain use the generic 
- [x] use virtual text for the description
- [x] adjust window size to care for virt_text

action_view


Ramilito:
@mosheavni Turned out really nice!

Issue #479, opened 2024-10-24
feat: debug container enhancement


Ramilito:
- [ ] Reattach to existing container
- [x] Create multiple debug containers (need to prompt for name)

Ramilito:
Skipping the reattach since that doesn't seem possible

PR #481, opened 2024-10-24
feat: enhanced debug for containers

Ramilito:
Fixes: #479 
- [x] use action_view
- [x] add options

action_view


Ramilito:
It doesn't refresh the view though, will create another pr for that

PR #484, opened 2024-10-25
Chore/remove create job from job

mosheavni:
it's not supported:
image

https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_job/#options


Ramilito:
nice catch!

Ramilito (approved):

PR #488, opened 2024-10-26
feat: confirm before deletion

kezhenxu94:
For deletion operations I believe users should be prompted to avoid mistakes

### DIFF
--- a/ftplugin/k8s_pods.lua
+++ b/ftplugin/k8s_pods.lua
@@ -58,23 +58,28 @@ local function set_keymaps(bufnr)
         end
       end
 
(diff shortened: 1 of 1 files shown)


Ramilito:
Hey! Thanks for taking time for this PR 🙏

While I agree with the pods, the port forward seems a bit too overly protective?

kezhenxu94:
> Hey! Thanks for taking time for this PR 🙏
> 
> While I agree with the pods, the port forward seems a bit too overly protective?

OK, reverted the one for PF

Ramilito (approved):

Ramilito:
LGTM! 🚀

PR #490, opened 2024-10-27
fix: random fixes

mosheavni:
random validations I added over time
change a bit the `not ready` in events
add `degraded` event


lua/kubectl/state.lua
```diff
@@ -89,8 +89,8 @@ function M.checkVersions()
   commands.shell_command_async("kubectl", { "version", "--output", "json" }, function(data)
     local result = decode(data)
     if result then
-      local clientVersion = result.clientVersion.gitVersion
-      local serverVersion = result.serverVersion.gitVersion
+      local clientVersion = result and result.clientVersion and result.clientVersion.gitVersion or "0.0"
```
Ramilito, line 92: Is the first check (result) needed? Don't we check that in the line above?

lua/kubectl/utils/events.lua
```diff
@@ -68,13 +66,16 @@ function M.ColorStatus(status)
     AlreadyMountedVolume = true,
     Available = true,
     ContainerCreating = true,
+    Degraded = true,
```
Ramilito, line 69: The not ready ones I understand but degraded only warning?

lua/kubectl/views/nodes/definition.lua
```diff
@@ -71,7 +71,7 @@ function M.getStatus(row)
   if ready and ready.status == "True" then
     return { symbol = hl.symbols.success, value = nodeConditions.NodeReady }
   end
-  return { symbol = events.ColorStatus("Error"), value = "Unknown" }
+  return { symbol = events.ColorStatus("NotReady"), value = "NotReady" }
```
Ramilito, line 74: I'm thinking we might keep the unknown as the last state and move the not ready into the if ready == true but in the elseif ready and ready.status == "false", what do you think?

lua/kubectl/state.lua
mosheavni: you are right

lua/kubectl/utils/events.lua
mosheavni: sure thing, moved it to errorStatuses

lua/kubectl/views/nodes/definition.lua
mosheavni: something like this?
[9cc3d80](https://github.com/Ramilito/kubectl.nvim/pull/490/commits/9cc3d809528c2ed5b92df63093e3054f1f66b425)

Ramilito (approved):

lua/kubectl/views/nodes/definition.lua
Ramilito: Exactly like that! Would have written code example but was on phone 😅

PR #495, opened 2024-10-28
feat: update pum position

mosheavni:
move pum window as we type
a little buggy but gives better experience


mosheavni:
fixed the bug with vim.schedule

Ramilito:
Hmm I don't super like this and dont think I've seen that anywhere, it's jumping around a lot

mosheavni:
Haha i thought you're gonna like it. Can we make it optional in the config? I for example see alot of benefit here

Ramilito:
Haha it's cool but not for me 😅, but sure put it behind an autocomplete config table 👍

mosheavni:
@Ramilito  made it optional. thanks!

Ramilito (approved):

Issue #485, opened 2024-10-25
convert port-forward float to action menu


mosheavni:
convert the port-forward float for pods and services to the new action menu with the virtual text.
right now it doesn't support adding more data (the containers to see their ports) so we need to adjust the `action_view` a bit

PR #500, opened 2024-11-04
fix: use action view for pf

Ramilito:
Fixes: #485 
Fixes: #501

- [x] make the pf command work for portforward, action_view doesn't support the local:container command merging
- [x] make action_view dynamic and only care about the data set instead of lines in buffer
- [x] make the enum tabbing work


mosheavni:
- [ ] map `<CR>` on ports to fill `container port: ` line

mosheavni:
port-forward case of merging the ports (i.e `8080:8080`) is a unique case, therefore, in the action_view callback we can just merge those manually, don't you think?
it's better than making something general and super complicated.

Ramilito:
> port-forward case of merging the ports (i.e `8080:8080`) is a unique case, therefore, in the action_view callback we can just merge those manually, don't you think? it's better than making something general and super complicated.

Not super sure yet, I think I have an idea on how to do this without making it super complicated but also not hardcoded, will give it a shot before returning to your suggestion

> also, why is --address localhost on a single item and not separated? 

Good question....

Ramilito:
Think I fixed all the issues now, did not understand this one though:  
> map <CR> on ports to fill container port:  line

mosheavni:
> Think I fixed all the issues now, did not understand this one though:
> 
> > map  on ports to fill container port:  line

like we have today, if you click enter on a port, it would fill the `container port` value

ooh I see it's not working anymore.
we can delay this to a separate issue

mosheavni:
when I click `<Tab>` on `address:` I get this:

```
E5108: Error executing lua: ...mavni/Repos/kubectl.nvim/lua/kubectl/resourcebuilder.lua:524: attempt to index a nil value
stack traceback:
        ...mavni/Repos/kubectl.nvim/lua/kubectl/resourcebuilder.lua:524: in function <...mavni/Repos/kubectl.nvim/lua/kubectl/resourcebuilder.lua:511>

```

mosheavni:
one more thing.
when the `type = 'flag'`, why do we even need to specify `options = {"true", "false"}`?
we can assume that if it's a flag, it's gonna be either `true` or `false`
is there a case where `flag` is not boolean?

Ramilito:
Thanks for testing it @mosheavni 🙏, missed those bugs.
Fixed the tab error and the extra space.

> one more thing. when the `type = 'flag'`, why do we even need to specify `options = {"true", "false"}`? we can assume that if it's a flag, it's gonna be either `true` or `false` is there a case where `flag` is not boolean?

You are correct I believe, we could probably refactor that into just using plain true and false boolean, leaving it in for this PR and we can do another one with the:

> like we have today, if you click enter on a port, it would fill the container port value

I think I understood that now, you can press enter on the container list? didn't even know we could do that and must have removed that code.

PR #508, opened 2024-11-17
feat/node-go-to-pods

mosheavni:

### DIFF
--- a/lua/kubectl/views/init.lua
+++ b/lua/kubectl/views/init.lua
@@ -197,37 +197,65 @@ function M.UserCmd(args)
   end)
 end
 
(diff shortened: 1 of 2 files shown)


lua/kubectl/views/init.lua
```diff
@@ -197,37 +197,74 @@ function M.UserCmd(args)
   end)
 end
 
-function M.set_and_open_pod_selector(name, ns)
```
Ramilito, line 200: I think we are using this method from other places right?

Ramilito:
Yeah I like this feature, the other day I was stuck copying node names and filtering on it.
LGTM 🚀!

Ramilito (approved):

PR #509, opened 2024-11-20
fix: set highlights as default

Saghen:
Setting the highlight as default allows color schemes to override them without depending on order of loading


Ramilito:
I wasn't aware of that property! Thank you for adding that 🙏

mosheavni (approved):

PR #518, opened 2024-11-27
fix/refresh aliases on ctx change

Ramilito:


mosheavni:
to update the window title upon cache loaded, I was thinking of firing an autocmd and registering autocmd listener to that event.

on `cache.lua`, change the call on `LoadFallbackData` to `M.load_cache()` to this:

```lua
    M.load_cache(M.cached_api_resources, function()
      -- fire autocmd event KubectlCacheLoaded
      vim.schedule(function()
        vim.cmd("doautocmd User KubectlCacheLoaded")
      end)
      if callback then
        callback()
      end
    end)
```

on `buffers.lua`, after creating the alias buffer, register:

```lua
  -- autocmd for KubectlCacheLoaded
  vim.api.nvim_create_autocmd("User", {
    pattern = "KubectlCacheLoaded",
    callback = function()
      vim.print("KubectlCacheLoaded")
      -- check if win and buf are valid
      local _, is_valid = pcall(vim.api.nvim_win_is_valid, win)
      local _, is_valid_buf = pcall(vim.api.nvim_buf_is_valid, buf)
      -- if both valid, update the window title
      if is_valid and is_valid_buf then
        -- ...
      end
    end,
  })
```

WDYT?

Ramilito:
@mosheavni found the issue, we were setting M.loading = false after the lineage feature was done, but if you have it turned off we never get there.
Moved it up and now it should work

mosheavni:
if I open the alias float before the cache was loaded, I see an incomplete number that never changes

Ramilito:
Sure thing!

mosheavni:
done, let me know what you think of the code :)

Ramilito:
Oh I see, you want to update it live and not when reopening it. I'll move the code up a step so it does the reload even for people that disable ctx

mosheavni (approved):

PR #529, opened 2024-12-17
chore/random-fixes

mosheavni:
1. informer had some errors, got rid of them and added some annotations
2. restore the table dashes to original dashes
3. add validation to definitions: crds, cronjobs, daemonsets, jobs. all based on real errors I got while working

### DIFF
--- a/lua/kubectl/utils/tables.lua
+++ b/lua/kubectl/utils/tables.lua
@@ -243,7 +243,7 @@ local function addDividerRow(divider, hints, marks)
     local count = divider.count or ""
     local filter = divider.filter or ""
     local info = resource .. count .. filter
-    local padding = string.rep("—", half_width - math.floor(#info / 2))
+    local padding = string.rep("-", half_width - math.floor(#info / 2))
     local selected = state.getSelections()
     local selected_count = vim.tbl_count(selected)
     if selected_count > 0 then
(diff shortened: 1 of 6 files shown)


lua/kubectl/utils/tables.lua
```diff
@@ -243,7 +243,7 @@ local function addDividerRow(divider, hints, marks)
     local count = divider.count or ""
     local filter = divider.filter or ""
     local info = resource .. count .. filter
-    local padding = string.rep("—", half_width - math.floor(#info / 2))
+    local padding = string.rep("-", half_width - math.floor(#info / 2))
```
Ramilito, line 246: Is this a different kind of unicode or a regular dash?

Ramilito (approved):

lua/kubectl/utils/tables.lua
mosheavni: the current one is some unicode special one which I thought was nice in https://github.com/Ramilito/kubectl.nvim/commit/e0bfb6eda2248e5f4ef83cf35cca44e95645e253#diff-c7fc12e395ddb005b4efffd49c7bf397d0b39048c10e57da1cf19b12b043287fR284

but it doesn't parse it as a ligature when I switched to Wezterm so I'm reverting to a simple plain dash (`-`)

PR #534, opened 2025-01-29
feat: added terminal_cmd to config, if set, exec in terminal and not a buffer

mbwilding:
Added terminal_cmd to config, if set, will exec in a standalone terminal of your choice, rather than a buffer.

### DIFF
--- a/lua/kubectl/actions/commands.lua
+++ b/lua/kubectl/actions/commands.lua
@@ -23,7 +23,7 @@ function M.configure_command(cmd, envs, args)
     if state.context["current-context"] then
       table.insert(result.args, "--context")
       table.insert(result.args, state.context["current-context"])
-      vim.print(result.args)
+      -- vim.print(result.args)
     end
   end
 
--- a/lua/kubectl/views/containers/init.lua
+++ b/lua/kubectl/views/containers/init.lua
@@ -22,14 +22,15 @@ function M.View(pod, ns)
 end
 
 function M.exec(pod, ns)
+  local args = { "exec", "-it", pod, "-n", ns, "-c ", M.selection, "--", "/bin/sh" }
   local cmd = "kubectl"
-  local args = string.format("exec -it %s -n %s -c %s -- /bin/sh", pod, ns, M.selection)
 
   if config.options.terminal_cmd then
-    vim.fn.jobstart(config.options.terminal_cmd .. " " .. cmd .. " " .. args)
+    local command = commands.configure_command(cmd, {}, args)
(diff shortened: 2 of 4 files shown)


lua/kubectl/actions/commands.lua
```diff
@@ -23,7 +23,7 @@ function M.configure_command(cmd, envs, args)
     if state.context["current-context"] then
       table.insert(result.args, "--context")
       table.insert(result.args, state.context["current-context"])
-      vim.print(result.args)
+      -- vim.print(result.args)
```
mbwilding, line 26: Figured this was accidentally left behind as a debugging measure.

Ramilito: Thank you I'll remove it in main right away 🙏

lua/kubectl/views/containers/init.lua
```diff
@@ -22,8 +22,15 @@ function M.View(pod, ns)
 end
 
 function M.exec(pod, ns)
-  buffers.floating_buffer("k8s_container_exec", "ssh " .. M.selection)
-  commands.execute_terminal("kubectl", { "exec", "-it", pod, "-n", ns, "-c ", M.selection, "--", "/bin/sh" })
+  local cmd = "kubectl"
+  local args = string.format("exec -it %s -n %s -c %s -- /bin/sh", pod, ns, M.selection)
+
+  if config.options.terminal_cmd then
+    vim.fn.jobstart(config.options.terminal_cmd .. " " .. cmd .. " " .. args)
```
Ramilito, line 29: We do some magic with the cmd inside the `commands.execute_terminal` (the `M.configure_command`), we keep it centrally since they are things we might want to do for every call (for example what context to run the command against)
Would this work for?
```lua
function M.exec(pod, ns)
  local cmd = "kubectl"
  local args = string.format("exec -it %s -n %s -c %s -- /bin/sh", pod, ns, M.selection)
  if config.options.terminal_cmd then
    cmd = config.options.terminal_cmd .. cmd
  end

  buffers.floating_buffer("k8s_container_exec", "ssh " .. M.selection)
  commands.execute_terminal(cmd, args)
end
```

mbwilding: Ok added that change to the PR, worked well with both code paths.

mbwilding:
Ready for review again, thanks for the feedback.

Ramilito:
Awesome feature and big thanks for the PR 🙏 

LGTM!

Ramilito (approved):

Issue #438, opened 2024-10-08
feat: ability to minimize floats and return to them


mosheavni:
this might require some refactoring, but in my day to day work, I exec into a pod, then I want to see the pod's yaml with `ge`, the I might wanna describe the pod,
then, I forgot something in the yaml and I want to return to it quickly, finally, I want to return to my exec float to continue working on the container.
right now, after every action I need to close the float/tab and open it again, invoking the commands over and over again.

Do you think it's easy to achieve?

Ramilito:
Discussed solution is a keymap to minimize (or quit act as a minimize), it should integrate with telescope and fzf-lua alternative is to have our own buffer selection only for our floats

PR #536, opened 2025-01-30
feat: recent view picker

Ramilito:
Fixes: #438 

- [x] Sort order  
- [ ] Exec not working -> moved to #539 
- [x] Event-handler is not working
- [ ] Explore integration with other pickers (fzf-lua) -> moved to #539 
- [x] Add binding to remove item from list

picker

What do you think, something like this? @mosheavni

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,10 @@ Processes kubectl outputs to enable vim-like navigation in a buffer for your clu
   <img src="https://github.com/user-attachments/assets/cb1f46be-fcc0-4a6d-9d1e-ffcd5bdb32b3" width="700px">
 </details>
 <details>
+  <summary>Picker for recent items</summary>
+  <img src="https://github.com/user-attachments/assets/fb519986-4d6e-437d-8fda-bd14199a71f2" width="700px">
+</details>
+<details>
   <summary>Lineage</summary>
   <sub>A plugin similar to <a href="https://github.com/tohjustin/kube-lineage/tree/master">kube-lineage</a></sub>
   <sub>⚠️ This is a beta feature and not all bugs are sorted out</sub>
--- a/ftplugin/k8s_events.lua
+++ b/ftplugin/k8s_events.lua
@@ -10,9 +10,9 @@ local function set_keymaps(bufnr)
     silent = true,
     desc = "View message",
     callback = function()
-      local message = tables.getCurrentSelection(unpack({ 7 }))
-      if message then
-        event_view.ShowMessage(message)
+      local ns, object, message = tables.getCurrentSelection(unpack({ 1, 5, 7 }))
+      if ns and object and message then
+        event_view.ShowMessage(ns, object, message)
       else
         print("Failed to extract event message.")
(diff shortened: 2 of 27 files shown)


Ramilito:
> oh wow looks excellent!

Ikr! Found an approach that didn't require too much work even, was dreading my original plan 😅 

> exec does not work, as already mentioned. 

Yeah, might do a second PR for that kind of deep nesting, it's probably best to change how we exec to allow it to work together with the picker

> we need a way to remove windows somehow, maybe the same as `kubectl_kill`?

Yeah that sounds good, `gd` as we delete the pods right?

mosheavni:
> > oh wow looks excellent!
> 
> Ikr! Found an approach that didn't require too much work even, was dreading my original plan 😅

honestly me too, I thought this is too much to code

> 
> > exec does not work, as already mentioned.
> 
> Yeah, might do a second PR for that kind of deep nesting, it's probably best to change how we exec to allow it to work together with the picker
> 
> > we need a way to remove windows somehow, maybe the same as `kubectl_kill`?
> 
> Yeah that sounds good, `gd` as we delete the pods right?

I think it's `gk` by default but anyhow it's `<Plug>(kubectl.kill)` as we do also on port-forwards

mosheavni:
> > oh wow looks excellent!
> 
> Ikr! Found an approach that didn't require too much work even, was dreading my original plan 😅

honestly me too, I thought this is too much to code

> 
> > exec does not work, as already mentioned.
> 
> Yeah, might do a second PR for that kind of deep nesting, it's probably best to change how we exec to allow it to work together with the picker
> 
> > we need a way to remove windows somehow, maybe the same as `kubectl_kill`?
> 
> Yeah that sounds good, `gd` as we delete the pods right?

I think it's `gk` by default but anyhow it's `<Plug>(kubectl.kill)` as we do also on port-forwards

mosheavni:
> > oh wow looks excellent!
> 
> Ikr! Found an approach that didn't require too much work even, was dreading my original plan 😅

honestly me too, I thought this is too much to code

> 
> > exec does not work, as already mentioned.
> 
> Yeah, might do a second PR for that kind of deep nesting, it's probably best to change how we exec to allow it to work together with the picker
> 
> > we need a way to remove windows somehow, maybe the same as `kubectl_kill`?
> 
> Yeah that sounds good, `gd` as we delete the pods right?

I think it's `gk` by default but anyhow it's `<Plug>(kubectl.kill)` as we do also on port-forwards

mosheavni:
feels amazing.
I'm not sure about including `namespace` and `filter` and `aliases` view on the menu because they have easy keymapping.
not sure about dropping or keeping it.

Ramilito:
No you are correct, excluding them now.
There is a problem with the sorting but otherwise I might merge it soon if we are happy with it?

mosheavni:
> No you are correct, excluding them now. There is a problem with the sorting but otherwise I might merge it soon if we are happy with it?

very happy.

Issue #537, opened 2025-01-31
Easy disabling of specific keymaps


jemag:
**Is your feature request related to a problem? Please describe.**
Currently, at least to my knowledge, there is no easy way to disable a specific keymap. If a keymap is overriden, then the mapping won't take place. However, if we do not want to have a mapping whatsover for a given action, there does not seem to be a way to simply disable it.

**Describe the solution you'd like**
Disable specific keymap (in my case, all the numbers `1`, `2`, etc.)

**Describe alternatives you've considered**
Tried using vim.keymap.del on `k8s_*` filetype but had issues with it, and it is also a painful way to do this, even if it worked.

**Additional context**
In my situation, most of the navigation I do in Neovim is through relative line numbers, so having those bounds to change view is a detriment for my workflow. I simply use `aliases` instead to access specific resources.

Ramilito:
Hey!

You can bind it to a <Nop> to remove the binding, the below should work, I'll try to find a better solution.

```lua
return {
  {
    "ramilito/kubectl.nvim",
    dir = "~/workspace/mine/kubectl.nvim/",
    config = function()
      require("kubectl").setup()

      local group = vim.api.nvim_create_augroup("kubectl_mappings", { clear = true })
      vim.api.nvim_create_autocmd("FileType", {
        group = group,
        pattern = "k8s_*",
        callback = function(ev)
          local k = vim.keymap.set
          local opts = { buffer = ev.buf }

          -- Global
          k("n", "<Nop>", "<Plug>(kubectl.view_deployments)", opts)
        end,
      })
    end,
  },
}
```

jemag:
Thank you for the information I should have thought of `<Nop>`. Although I do think it would be nice to have an easier mechanism.

With the current setup, users have to find the equivalent action of each key for before unbinding it. 
In the example above, the user has to look "what is the current action for the key `1`", then unbind that. I am not quite sure also what would happen if that action is used for multiple keybinds (i.e. if `<Plug>(kubectl.view_deployments)` add 2 keymaps, then doing this I am guessing would remove both).

Not sure what the best approach is, but it would be nice to simply identify keybinds to unbind directly. Some plugins, for example, provide a list of mapping that can be mapped to actions or nothing:
```lua
    mappings = {
      ["a"] = "add",
      ["d"] = "delete",
      ["r"] = "rename",
      ["/"] = "none",
      ["?"] = "none",
      ["z"] = "none",
    },

```

[...]

PR #538, opened 2025-02-02
fix: revamp keybinding setup

Ramilito:
Motivation: using ftplugin files to add mappings makes it very difficult to handle the order since there is just one event to attach to. This has forced us to add defers with timings to do things later and the limitation that we cannot override plug functions.

Solution: use regular files that exports a keymap table that we can merge with the global one at file open. This is still a ftplugin event but we have everything in one table before applying it.
 
Fixes: #545, #537

- [x] Use vim methods to check for existing mappings
- [x] Allow views to override default mappings
- [x] Fix mappings not visible on first load in floats 
- [x] Remove gK mapping in pods
- [x] Allow users to unmap bindings (just need to update readme) 
- [x] Migrate all current mappings into new way
- [x] Migrate action view

Bonus:
- [x] Fix fallback bug introduces in previous patch
- [x] Fix crds and api-resource select not working
- [x] incorrect validation when cordoning nodes fix 
- [x] Apply a sorting to the hints


mosheavni (approved):
LGTM

PR #544, opened 2025-02-07
feat: delete multiple selected resources

Ramilito:
Fixes: #541 

This PR makes the float better looking and more informative on what resources we are acting on.
This PR also adds the selection to the delete resources command, not just the kill command 

I feel the `gk` command is obsolete with the `gD` command @mosheavni, do you remember why we use two similar commands to do the same thing?


mosheavni:
No real reason, can be unified with gD

Issue #555, opened 2025-02-15
Divide UI between interactive and non-interactive


jemag:
**Is your feature request related to a problem? Please describe.**

Currently, kubectl.nvim places the cursor within a buffer which only a portion can be interacted with.
For example:
If I load the tool, and want to select the first pod in my list, I need to bring the cursor down 7 lines before being able to act on something interactive:
![Image](https://github.com/user-attachments/assets/9a24cde7-5534-4432-a980-f714ff41232e)

**Describe the solution you'd like**
I believe that for the sake of efficiency and seamless interaction, it would probably be better to divide the UI between the interactive and the non-interactive section. I am not sure what would be the best/easiest way to accomplish this, perhaps there could simply be 2 windows:
- the top one containing the header and resource title
- the bottom one where you can interact with the actual resources of the cluster
E.g. separate the green and red section here:
![Image](https://github.com/user-attachments/assets/6da27d43-c491-490a-afbd-218622af1ae2)
Another advantage is that for long list of resources, your users could still be able to see the top header while scrolling down.

With this workflow, once the user triggers kubectl.nvim, cursor would automatically be placed within the interactive section (i.e., first resource in the list), and the user can start acting directly.

**Describe alternatives you've considered**
Keep things as is

Ramilito:
Yeah that's a solid suggestion, I'm playing around with a way of implementing it here #556, lets see where that takes us

PR #556, opened 2025-02-16
feat: use split buffer for header management

Ramilito:
Fixes: #555 
Have the header in its own buffer

- [x] prevent manual resizing (by resizing on draw)
- [x] add auto rezie when toggle
- [x] print divider in winbar
- [x] header is preventing navigation if you try to jump over the header - REJECTED, lets allow navigating to the header since it's an intentional action.
- [x] winbar bg color should be same as normal
- [x] close the header when closing main buffer
- [x] prevent navigation to the header
- [x] Float divider missing
- [x] make sure different header configs works

Status bar: Not explored, this is the statusbar of the header buffer, what should we put there?
winbar: Use as divider, this is because we cannot remove it when having splits so we might as well use that row for the divider (looks pretty nice)


lua/kubectl/resourcebuilder.lua
```diff
@@ -282,7 +285,20 @@ function ResourceBuilder:setContent(cancellationToken)
...
+      buffers.set_content(self.buf_header_nr, { content = {}, marks = {}, header = self.header })
+      buffers.set_content(self.buf_nr, { content = self.prettyData, marks = self.extmarks, header = {} })
+      vim.schedule(function()
+        vim.api.nvim_set_option_value("winbar", self.header.divider, { scope = "local", win = self.win_nr })
+        vim.api.nvim_set_option_value("winbar", "", { scope = "local", win = self.win_header_nr })
+        vim.api.nvim_set_option_value("statusline", " ", { scope = "local", win = self.win_header_nr })
+      end)
+    end
+  else
+    buffers.set_content(self.buf_nr, { content = self.prettyData, marks = self.extmarks, header = self.header })
```
Ramilito, line 300: Make better

PR #566, opened 2025-03-02
WIP - v2.0

Ramilito:
Moved this PR into a new one since this is stating to lag quite a lot: #572


Ramilito:
Preliminary performance testing, the get_table does gets from the cache, sorts and processes the data in ~5ms (notice it ran twice during the profiling)

image

mosheavni:
exciting and interesting stuff! maybe i'll have to learn rust now :)

Ramilito:
> exciting and interesting stuff! maybe i'll have to learn rust now :)

Yeah I still don't feel like I know Rust, but it's fun!

With the kube.rs client doing all the types for kubernetes it gets so much easier to write things for it, sorting, filtering or just in general manipulation of the data, it just works!

Not to mention the informer!

I have @Saghen to thank for the rust integration, was looking at blink.cmp source code and got super excited when I saw that this was possible, Thank you @Saghen 🙏

mosheavni:
@Ramilito I think if the informer's code is converted to rust the performance and stability would climb

Ramilito:
> @Ramilito I think if the informer's code is converted to rust the performance and stability would climb

Exactly, I've already noticed that it's working so much better. There is even a better way to work with that but I couldn't get it to work. It's called a reflector, it basically combines the informer and the database. That would make our state super stable and performant.

Ramilito:
> > I have @Saghen to thank for the rust integration, was looking at blink.cmp source code and got super excited when I saw that this was possible, Thank you @Saghen 🙏
> 
> Wow this looks awesome, congrats! Happy to see the rust worm spreading :D This encouraged me to throw up [a template](https://github.com/Saghen/neovim-lua-rust-template/tree/main) for others as well. Would you want to use the downloader code for prebuilt binaries as well? I've been thinking about moving that to a separate library for others to use

Yeah my other Kubernetes tools ([kubesess](github.com/ramilito/kubesess) and [kubediff] github.com/ramilito/kubesess) are also written in Rust so this was the missing link 😄 

That download file was what made me look into blink.cmp, I had seen the build notification and wanted a closer look and now my whole weekend went this PR 😅. 
I wasn't a big fan of the complexity though, I guess if there is no better way then it can't be helped but is there no way to build it and just push the artifact into the repo? I'm thinking that the package managers will clone the repo and then it would be there?

Saghen:
> is there no way to build it and just push the artifact into the repo?

You could but given a binary size of 3MBs with 8 platforms (how many blink.cmp supports atm), you end up increasing the size of the repo by 24MBs on every release. Git LFS could work if package managers supported it but iirc Github charges for storage and bandwidth with that.

> I wasn't a big fan of the complexity though

Yeah, for blink.cmp, I wanted to make it as seamless as possible since there's many new users using it. But I could setup a much simpler version of the downloader that simply downloads the binary for the current tag when called via `build = function() require('kubectl').download_prebuilt() end` in a lazy definition. This wouldn't perform checksum, out of date per commit, or any other checks, but that should be fine in 99% of cases.

Ramilito:
> > is there no way to build it and just push the artifact into the repo?
> 
> You could but given a binary size of 3MBs with 8 platforms (how many blink.cmp supports atm), you end up increasing the size of the repo by 24MBs on every release. Git LFS could work if package managers supported it but iirc Github charges for storage and bandwidth with that.
> 
> > I wasn't a big fan of the complexity though
> 
> Yeah, for blink.cmp, I wanted to make it as seamless as possible since there's many new users using it. But I could setup a much simpler version of the downloader that simply downloads the binary for the current tag when called via `build = function() require('kubectl').download_prebuilt() end` in a lazy definition. This wouldn't perform checksum, out of date per commit, or any other checks, but that should be fine in 99% of cases.


I see what you mean, and I just checked the size of the binary, it's about 8mb per binary 😅...that's a bit too much for a git clone.
If you make it a separate library for the download file I would love to use that! If everyone helps out then I guess the fully working one wont be too difficult to maintain. 
Thanks and sorry for sounding like a choosing beggar 😃

mosheavni:
hi @Ramilito  I follow this branch (I build with `cargo build --release`)
there are some issues but you're still working on it, I'm wondering if you want me to share my findings.

Ramilito:
Please do!

Ramilito:
@mosheavni, it's getting pretty difficult to keep both solutions live at the same time, I'm going to start not supporting the old solution in this branch, which will make it quite difficult for you to use this branch 😞

Ramilito:
> I've thrown together a simpler version of the [downloader in a separate repo](https://github.com/Saghen/blink.download) and included an example of how to use it. I've also updated the [neovim-lua-rust-template](https://github.com/Saghen/neovim-lua-rust-template) to include CI for release builds and an example for integrating the downloader. I have yet to test all of this though, so ymmv, but I'll lyk as soon as I've tested it end to end.

That's great! I'll try it out soon and let you know how it went 🙏

Saghen:
Just dogfooded it with https://github.com/Saghen/blink.pairs and it seems to be working well! Quite a few updates to the template to get it all working. Don't hesitate to ping if you run into any issues!

Ramilito:
> Just dogfooded it with https://github.com/Saghen/blink.pairs and it seems to be working well! Quite a few updates to the template to get it all working. Don't hesitate to ping if you run into any issues!

Thank you, will do that, so far the struggles are a bit contextual to my setup, needed another nesting of FFI to reuse code from `kubectl/describe.go`...tried rewriting it but 5k code is rough to translate 😓 .

So I have to build a Go library as well as a Rust one 😄.


The only issue so far from reusing your code is was an error from the `cross.toml` zig spec.
image

Ramilito:
@Saghen Will test some more tomorrow but I did manage to get it to download the binaries (55mb or something, thinking of maybe doing some zip/unzip shenanigans). 
Took me too long than I'd like to admit to figure out that I needed `tags` to get it to work 😅.
 
But it did not play well with the  `kubectl/client/rust/init.lua` file, the one that sets the cpath. I had to remove it for the `blink.download` to download the binaries, which it did!

There was some other error later with the `client.implementation` call, will dig into it more.

Ramilito:
@Saghen Got it working! The error was how I initiate the client where I was not importing the actual binaries (since that was done in the cpath setting code previously but replaced by `blink.download`), you did write that we only needed to require the module but it was late and I missed that 😅

Saghen:
> But it did not play well with the kubectl/client/rust/init.lua file, the one that sets the cpath

Oh that's weird, I'm [using that on blink.pairs](https://github.com/Saghen/blink.pairs/blob/main/lua/blink/pairs/rust.lua) so that blink.download is an optional dependency

> thinking of maybe doing some zip/unzip shenanigans

Yeah, we should support this natively in blink.download, I'd imagine gzip would be the only safe bet. Also need to think how to make the downloader flexible enough such that post-download transforms are easy

Ramilito:
> Oh that's weird, I'm [using that on blink.pairs](https://github.com/Saghen/blink.pairs/blob/main/lua/blink/pairs/rust.lua) so that blink.download is an optional dependency

Not sure what happened but it did work later, I'm guessing I had a mix of git tag + local dev env, I'll let you know if I can recreate it.
 
> Yeah, we should support this natively in blink.download, I'd imagine gzip would be the only safe bet. Files will always be written as `'lib' .. binary_name .. get_lib_extension()` so it's not easy to write this as-is, I'll look into making that more flexible

That would be awesome, I'm overwhelmed with this PR currently but later on I might be able to assist 👍

lua/kubectl/views/clusterrolebindings/init.lua
```diff
@@ -0,0 +1,62 @@
...
+
+local resource = "clusterrolebindings"
+
+---@class Module
+local M = {
+  definition = {
+    resource = resource,
+    display_name = string.upper(resource),
+    ft = "k8s_" .. resource,
+    gvk = { g = "rbac.authorization.k8s.io", v = "v1", k = "clusterrolebinding" },
```
mosheavni: what if someone has a cluster with v1beta1? i think this should be handled dynamically

Ramilito: This is how it looks like in main, so we are not loosing any functionality with this code, although, we should have a different discussion on this since there is multiple ways to handle this.
For the core api's, I think we are very safe with the v1 currently

image

mosheavni:
hey @Ramilito  i really would like to see the progress, building is not working anymore.
can you describe the steps to run this branch?
thanks.

Issue #569, opened 2025-03-17
feat: notify from ffi


Ramilito:
**Is your feature request related to a problem? Please describe.**

Currently we don't have an option to return updates from the Rust code and need to use a polling mechanism instead.

**Describe the solution you'd like**

I'd like the Rust code to be able to push changes as a notification instead of using polling.

**Describe alternatives you've considered**

**Additional context**

https://github.com/mlua-rs/mlua/issues/76
https://github.com/mlua-rs/mlua/issues/143

Issue #568, opened 2025-03-17
feat: tailored describe using the Go code


Ramilito:
**Is your feature request related to a problem? Please describe.**

The Go binary is adding 60Mb to the binary because it includes it's own client and a lot of other things.

**Describe the solution you'd like**

I'd like to only use the printer to avoid re-inventing the describe solution but exclude as much other stuff as possible

**Describe alternatives you've considered**

Doing our own is a lot of work, tried that approach but not worth it.

**Additional context**

Spinoff from: #566

Ramilito:
Gone through a lot of options for this one and all are dead ends. 
The most promising one is to gzip the dyslib produced but that would force the implementation of a unzipping step from the user, which isn't needed for now.

If someone complains about the binary size, we could do this and have it as an optional setup.

PR #572, opened 2025-04-21
Release 2.0

### LINKED ISSUE #530: bug: refresh not working properly
It seems that we are not handling all refresh properly, not sure where the bug is or how to reproduce it but some people are experiencing that they have to `gr` manually to get an updated view

- [ ] Opening kubectl.nvim into a fallback view doesn't start the informer
- [ ] Add event when table is empty throws error

Ramilito:
This PR intends to move a lot of our logic into rust and to use kube.rs as a client to communicate to the kubernetes API.
The goal is to lessen or even remove the curl and perhaps even the kubectl dependency. Also, to remove our own implementation of the informer.

Fixes #530

- [x] Implement informer through kube.rs (called watcher)
- [x] Do sorting in rust
- [x] Use DB cache in rust
- [x] Do the processing of rows in rust
- [ ] When all edge cases are solved, do this for all views
     - [x] Deployment
     - [x] action (supports both kube.rs and kubectl)
     - [x] clusterrolebinding
     - [x] configmap
     - [x] contexts
     - [x] crds
     - [x] cronjobs
     - [x] daemonset
     - [x] deploymen
     - [x] events
     - [x] fallback
     - [x] filter
     - [x] hpa
     - [x] ingresses
     - [x] jobs
     - [x] lineage
     - [x] nodes
     - [x] pod_logs
     - [x] pods
     - [x] port_forw
     - [x] pv
     - [x] pvc
     - [x] replicase
     - [x] sa
     - [x] secrets
     - [x] services
     - [x] storageclass
     - [x] statefulset
     - [x] secret_de
     - [ ] namespace
     - [ ] helm
     - [ ] overview
     - [ ] container (kubectl exec and debug functionality needs migrating)
     - [ ] top_pods
     - [ ] top_nodes
- [x] Distribution of binary
- [x] Optimize get_table (make async)
- [x] FIlter
- [x] Filter_label
- [x] Sorting
- [x] Get resource yaml
- [x] Describe resource
- [x] Edit resource
- [x] Port forward start
- [x] Port forward list in view and show icon on resource
- [x] Port forward stop
- [x] Logs
- [x] Scale 
- [x] Fallback/crds
- [x] Restart
- [x] Negative filter
- [ ] Describe resource live update
- [x] Watch resource 
- [ ] Select and navigate to child resources (deploy -> pod filtered)
- [x] Cordone, uncordone, drain

Bonus:
- [x] Memory leak caused by extmarks
- [x] Builder reference issue causing weird state issues when multiple views open at the same time

Moved tasks:
- [ ] Slimmer go package #568  
- [ ] Notify main lua thread instead of a loop #569

This is so far at least 2.5x faster than the pure lua implementation.

### Benchmark
This is a test taken over 1min (2 sec redraw) resulting in 1696ms (aggregated all the timings that are not in the Rust one) for the Lua implementation vs 700ms(rounded up since it only ran 29 times vs lua 31 times) for the Rust one:

image
image

Also did this test for 10min and it scales linearly, so went to 17 000ms compared to 7000 ms for 300 runs.


Ramilito:
Moving this into a correctly named branch + pr name for smoother releases

PR #583, opened 2025-06-30
Feat/exec new tab

mosheavni:


lua/kubectl/resources/containers/init.lua
```diff
@@ -64,6 +64,10 @@ local function attach_session(sess, buf, win)
   vim.cmd.startinsert()
 
   local timer = vim.uv.new_timer()
+  if not timer then
```
Ramilito, line 67: Did this give you an error anytime? I usually don't do error handling on builtin functionality that is supposed to be managed by neovim

PR #613, opened 2025-07-24
fix: add missing view identifiers in viewsTable

jemag:
Just a few missing identifiers that would lead to default view instead of the custom one.


Ramilito:
Could you just fix this first:
image

jemag:
my bad, didn't realize you had your own stylua config; usually I try to avoid formatting otherwise. Should be all good.

Ramilito:
No worries! Looking good!

PR #619, opened 2025-07-28
feat: use-kubectl-bufname-prefix

mosheavni:
This PR standardizes buffer naming by adding a `kubectl://` prefix to all kubectl
plugin buffers.

Changes

- Buffer naming consistency: All kubectl buffers now use the `kubectl://`
prefix for better identification and organization
- Simplified internal buffer names: Removed redundant kubectl_prefixes from
internal buffer names (e.g., kubectl_filter → filter, kubectl_confirmation →
confirmation)
- Updated buffer lookup logic: Modified get_windows_by_name() to properly
find buffers with the new naming scheme

Benefits

- Clearer buffer identification in buffer lists
- Consistent naming pattern across all kubectl buffers
- Better namespace separation from other plugin buffers


Ramilito:
I like the idea but I think this needs more work, we use the buffername to find the resource and views.
For example when we call:
```lua
        local _, buf_name = pcall(vim.api.nvim_buf_get_var, 0, "buf_name")
        local view = require("kubectl.views").resource_and_definition(vim.trim(buf_name))
```
We don't go through: `get_windows_by_name` and I think there are other places that might need work

mosheavni:
> I like the idea but I think this needs more work, we use the buffername to find the resource and views. For example when we call:
> 
> ```lua
>         local _, buf_name = pcall(vim.api.nvim_buf_get_var, 0, "buf_name")
>         local view = require("kubectl.views").resource_and_definition(vim.trim(buf_name))
> ```
> 
> We don't go through: `get_windows_by_name` and I think there are other places that might need work

it's a non issue.

on my branch if I run `=vim.api.nvim_buf_get_var(0, "buf_name")` on pods the output is simply `pods`
Jul-29-2025 09-40-59

we set buf a variable called `buf_name`:
https://github.com/Ramilito/kubectl.nvim/blob/92ab9baa43cb37a8c0b3ed69faea7f7cc1019534/lua/kubectl/actions/layout.lua#L23-L23

and set the bufname:
https://github.com/Ramilito/kubectl.nvim/blob/92ab9baa43cb37a8c0b3ed69faea7f7cc1019534/lua/kubectl/actions/buffers.lua#L12-L12

Ramilito:
You are correct, approved!

PR #628, opened 2025-08-03
chore: remove-unused-func

mosheavni:
remove unused function.
i wonder if there's an efficient way of knowing which functions are unused.


Ramilito:
Not that I'm aware of :(, that's the issue with these weak typed languages

PR #631, opened 2025-08-03
feat/auto-select-existing-sess-labels

mosheavni:
1. open filter label menu
2. add a custom label
3. uncheck it
4. check some existing label
5. hit `y` to accept
6. open filter label float again
7. now the label from step 2 is unchecked and the label from step 4 is checked


Ramilito:
Nice!

PR #643, opened 2025-08-13
fix/containers-loop

### LINKED ISSUE #642: container float not refreshing
currently the containers float (after clicking select on a pod) is not auto refreshing.
Right now, I have to close and reopen for it to refresh. any way to add it to the loop?
thanks.

mosheavni:
add loop for containers float. fixes #642


Ramilito:
Lets skip this one in favor of #645

Issue #584, opened 2025-07-03
feat: smarter live describe view


Ramilito:
The current describe view has a rather stupid loop that just makes calls to fetch the describe again.
Maybe we can add a toggle? similar to what k9s does?

Ramilito:
The stupid loop is causing other issues such as #635.

PR #645, opened 2025-08-14
feat: notify from ffi

Ramilito:
Fixes: #569 and #642

it also enables us to make a better: #584

### DIFF
--- a/lua/kubectl/event_queue.lua
+++ b/lua/kubectl/event_queue.lua
@@ -25,14 +25,14 @@ function M.unregister(name)
 end
 
 function M.start(interval_ms)
-  client.setup()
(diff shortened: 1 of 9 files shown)


lua/kubectl/event_queue.lua
```diff
@@ -18,6 +18,13 @@ function M.register(name, fn_or_code)
     error("callback must be a function or a code string that returns a function")
   end
   M._callbacks[name] = fn
+
+  vim.api.nvim_create_autocmd({ "QuitPre", "BufHidden", "BufUnload", "BufDelete" }, {
```
mosheavni, line 22: better register autocmds with an augroup

PR #654, opened 2025-08-30
fix: fix incorrect view mapping in README

YangKian:
Hi, thanks for the great work. I noticed that the `kubectl.view_pv` and `kubectl.view_pvc` shortcuts weren't working after I configured them according to the README. It seems the correct commands are `kubectl.view_persistentvolumes` and `kubectl.view_persistentvolumeclaims`. This PR fixes the problem.


Ramilito:
Nice catch! 🚀

PR #658, opened 2025-09-07
fix: better portforward stability

Ramilito:
I've noticed that Portforward socket is being closed without notice, this PR aims to solve that


Ramilito:
It's looking better but will use it for a while to make sure

PR #662, opened 2025-09-23
fix: don't hang on bad handle_cluster_unavailable

Ramilito:
Currently, when we target a non responsive cluster the current neovim session gets hanged and you have to close the terminal and leaves a zombie process alive.
This PR makes this not happen and handles multiple tries to a faulty api.


Ramilito:
@mosheavni Could you test this out with you VPN? It's working quite good but I'm guessing a bit on the some of the code 😅

PR #664, opened 2025-09-23
fix: update key mapping for ServiceAccounts in init.lua and README

YangKian:
Hello, I found `kubectl.view_sa` doesn't exist, this pr fix this problem


Ramilito:
Thank you 🙏

PR #680, opened 2025-12-19
Add Socks5 support to Kube

mbwilding:
Additionally I fixed a dead code warning.

Everytime I update through lazy I have to rebuild manually with `cargo build --release --features kube/socks5`.
I added a build command to lazy on my end but it times out as it takes to long.

### DIFF
--- a/kubectl-client/Cargo.toml
+++ b/kubectl-client/Cargo.toml
@@ -3,8 +3,8 @@ name = "kubectl-client"
 version = "0.1.0"
 edition = "2021"
 
+
 [features]
-default = ["kube/socks5"]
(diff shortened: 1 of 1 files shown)


Ramilito:
LGTM 🚀!

Thanks for sharing 🙏

mbwilding:
Thanks for reviewing and merging.

Any chance to push a tag to set off a release? Happy to wait for your next release cycle though.

Ramilito:
Ah, we usually have an automatic trigger for release but we need to decorate the commit with a keyword.
I've triggered one manually now but for future reference:

image

Issue #510, opened 2024-11-20
feat: blink.cmp source


saghen:
**Is your feature request related to a problem? Please describe.**
The current completion system uses the built-in completions and does not support external completion engines

**Describe the solution you'd like**
Support for external completion engines such as blink.cmp and/or nvim-cmp

**Additional context**
Hey there, love the plugin! It's getting me to finally switch off my shell filled with aliases :smile: I'm the author of [blink.cmp](https://github.com/Saghen/blink.cmp) and was curious if you'd accept a built-in source for it. I believe I'll need to make some changes on blink's side which I'm more than happy to do.

Ramilito:
> Hey there, love the plugin! It's getting me to finally switch off my shell filled with aliases 😄

I'm in the same boat as you, think this plugin turned out really good!
Seen `blink.cmp` around but never had the time to test it out, just did and liking it a lot!

Would the end results be that we have an optional dependency to an external cmp engine but if not, we fallback on the built-in one?

I have not looked at the code for cmp sources so not sure what changes needs to be done on our side but I like the idea and let me know if you need me to do anything!

mosheavni:
i think it's wise.
if you plan on implementing it, i think it must come with support for the most popular (nvim-cmp) and yours at the same time.

Ramilito:
> i think it's wise. if you plan on implementing it, i think it must come with support for the most popular (nvim-cmp) and yours at the same time.

If it's an optional dep with fallback on the current built-in I would say it's enough to make the implementation extensible so that nvim-cmp can be added separately

PR #699, opened 2026-01-03
feat: add lsp support

Ramilito:
Fixes: #510


Ramilito:
Hey @saghen! Finally in a place where I can work on #510, could you guide me? Any examples I can follow and maybe a way for me to test it locally?

saghen:
Hey! If you haven't already seen it, there's a [source boilerplate](https://cmp.saghen.dev/development/source-boilerplate.html) which goes over most of the features. For testing locally, I use `lazy.nvim`'s `dev = true` and then add my source to blink.cmp like

```lua
require('blink.cmp').setup({
  sources = {
    per_filetype = { your_filetype = { 'kubectl', inherit_defaults = true } },
    providers = {
      kubectl = { module = 'kubectl.completion.blink' }
    }
  }
})
```

You can however implement this exact configuration programmatically (which I'd recommend) so your users don't need to configure this themselves.

```lua
local cmp = require('blink.cmp')
cmp.add_source_provider('kubectl', { module = 'kubectl.completion.blink' })
cmp.add_filetype_source('your-filetype', 'kubectl')
```

Fyi, even better than this, you can implement an LSP so that all completion engines, including the built-in, work with one implementation. It's a bit trickier though. Here's [obsidian's implementation](https://github.com/obsidian-nvim/obsidian.nvim/tree/main/lua/obsidian/lsp), [crate.nvim's implementation](https://github.com/saecki/crates.nvim/blob/main/lua/crates/lsp.lua), and [a WIP PR](https://github.com/neovim/neovim/pull/24338) which you can copy to make the interface a bit nicer.

Ramilito:
Thank you for the guidance! I went with the LSP approach and it just automatically works with Blink.cmp, amazing!

I'm going to put som efforts on the actual stuff returned but so far it looks great!
image

PR #713, opened 2026-01-09
feat: toggle and reorder columns

### LINKED ISSUE #712: feat: toggle columns
right now the columns on every view are static.
maybe we can dynamicaly let users hide them?
for every view, I would open a new float that displays boxes for each column, for example for pod:

- [X] NAMESPACE
- [X] NAME
- [X] READY
- [X] STATUS
- [X] RESTARTS
- [X] CPU
- [X] MEM
- [X] %CPU/R
- [X] %CPU/L
- [X] %MEM/R
- [X] %MEM/L
- [X] IP
- [X] NODE
- [X] AGE    

and allow to toggle the columns.
this would also allow us to add even more columns that are hidden by default but if the user wants them, he could toggle them in.

mosheavni:
columns

ability to change the columns displayed
and their order
persist on disk - per view.
closes #712

### DIFF
--- a/lua/kubectl/views/columns/mappings.lua
+++ b/lua/kubectl/views/columns/mappings.lua
@@ -78,5 +78,4 @@ function M.register()
   mappings.map_if_plug_not_set("n", "R", "<Plug>(kubectl.reset_order)")
 end
 
-
 return M


lua/kubectl/actions/buffers.lua
```diff
@@ -127,12 +127,12 @@ function M.apply_marks(bufnr, marks, header)
   end
 
   local ns_id = api.nvim_create_namespace("__kubectl_views")
-  pcall(api.nvim_buf_clear_namespace, bufnr, ns_id, 0, -1)
-
   local header_data = header and header.data
   local header_row_offset = header_data and #header_data or 0
 
   vim.schedule(function()
+    pcall(api.nvim_buf_clear_namespace, bufnr, ns_id, 0, -1)
```
Ramilito: This introduces a bug in rendering! Don't remember the exact bug but it messes up the marks, I think the tab selection stops working amongst other things

lua/kubectl/views/columns/definition.lua
```diff
@@ -0,0 +1,20 @@
...
+
+M.definition = {
+  resource = "columns",
+  display = "Toggle Columns",
+  ft = "k8s_columns",
+  title = "Columns",
+  hints = {
+    { key = "<Plug>(kubectl.tab)", desc = "toggle" },
+    { key = "<Plug>(kubectl.move_up)", desc = "move up" },
+    { key = "<Plug>(kubectl.move_down)", desc = "move down" },
```
Ramilito: Maybe we can simplify the code and just use the buffer as is? like the order of the items, we could add an ID to identify the columns or just name match them later?

lua/kubectl/actions/buffers.lua
mosheavni: oh really? when it's above the vim.schedule then the hints extmarks are rendered weird.
I.E:
image

PR #730, opened 2026-01-22
Fix: Pass namespace to reflector for namespace-scoped RBAC users

### LINKED ISSUE #729: Title: Reflector ignores namespace config, always queries cluster-scope
Describe the bug

  When using kubectl.nvim with a Kubernetes user that has namespace-scoped
  permissions only (no cluster-admin), the plugin fails to display any
  resources (pods, deployments, etc.) even though namespace and
  namespace_fallback are correctly configured.

  The root cause is that start_reflector_async is called with hardcoded
  namespace = nil, causing the Rust watcher to make cluster-scope API requests
  instead of namespace-scoped ones. This is inconsistent with draw() which
  correctly uses state.ns.

  To Reproduce

  1. Configure kubectl with a user that has RoleBinding to a single namespace
  only (no ClusterRoleBinding)
  2. Add plugin config with explicit namespace:
  require("kubectl").setup({
      namespace = "my-namespace",
      namespace_fallback = { "my-namespace" },
  })
  3. Open kubectl.nvim with <leader>k
  4. See empty resource list
  5. Check ~/.local/state/nvim/kubectl.log - see 403 errors with "at the
  cluster scope"

  Expected behavior

  When namespace is set in config, the reflector should query that namespace
  only, respecting the user's limited permissions. Resources should be
  displayed correctly.

  Screenshots

  Log output from ~/.local/state/nvim/kubectl.log:
  WARN kube_runtime::watcher: watch list error with 403: Api(ErrorResponse {
    status: "Failure",
    message: "pods is forbidden: User \"my-user\" cannot list resource \"pods\"
   in API group \"\" at the cluster scope",
    reason: "Forbidden",
    code: 403
  })

  Environment:
  - OS: Linux (Ubuntu)
  - Neovim version: 0.10+
  - kubectl.nvim version: 2.x (with Rust client)

  Additional context

  The issue is in two files where namespace = nil is hardcoded:
  - lua/kubectl/resource_factory.lua:265
  - lua/kubectl/resources/fallback/init.lua:44

  Meanwhile, draw() in resource_factory.lua:282 correctly uses:
  local namespace = (state.ns and state.ns ~= "All") and state.ns or nil

  This affects all users working in multi-tenant Kubernetes clusters with
  namespace-scoped RBAC permissions (very common setup).

BioWare:
### Problem

Users with namespace-scoped Kubernetes permissions (no cluster-admin) see  empty resource lists because `start_reflector_async` always uses `namespace = nil`, causing cluster-scope API requests that return 403 Forbidden.

### Solution

Pass `state.ns` to the reflector, consistent with how `draw()` already handles it:
- If `state.ns == "All"` → pass `nil` (cluster-scope, unchanged behavior)
- If `state.ns == "my-namespace"` → pass the namespace (namespace-scope)

### Changes

- `lua/kubectl/resource_factory.lua`: Use `state.ns` in `view()`
- `lua/kubectl/resources/fallback/init.lua`: Same fix for fallback resources

### Impact

- **Backward compatible**: Users with cluster-admin permissions see no change
- **Fixes**: Namespace-scoped users can now use the plugin by setting  `namespace` in config

### Testing

Tested with a user that has RoleBinding to a single namespace only:
- Before: Empty pod list, 403 errors in `~/.local/state/nvim/kubectl.log`
- After: Pods displayed correctly

Closes #729

### DIFF
--- a/lua/kubectl/resource_factory.lua
+++ b/lua/kubectl/resource_factory.lua
@@ -262,7 +262,9 @@ function M.new(resource)
     builder.buf_nr, builder.win_nr = buffers.buffer(definition.ft, builder.resource)
     state.addToHistory(builder.resource)
 
-    commands.run_async("start_reflector_async", { gvk = definition.gvk, namespace = nil }, function(_, err)
+    -- Use namespace from state to support users with namespace-scoped permissions only
+    local ns = (state.ns and state.ns ~= "All") and state.ns or nil
+    commands.run_async("start_reflector_async", { gvk = definition.gvk, namespace = ns }, function(_, err)
       if err then
         return
       end
--- a/lua/kubectl/resources/fallback/init.lua
+++ b/lua/kubectl/resources/fallback/init.lua
@@ -41,7 +41,9 @@ function M.View(cancellationToken, kind)
   builder.definition = M.definition
   builder.buf_nr, builder.win_nr = buffers.buffer(builder.definition.ft, resource.name)
 
-  commands.run_async("start_reflector_async", { gvk = M.definition.gvk, namespace = nil }, function()
+  -- Use namespace from state to support users with namespace-scoped permissions only
+  local ns = (state.ns and state.ns ~= "All") and state.ns or nil
+  commands.run_async("start_reflector_async", { gvk = M.definition.gvk, namespace = ns }, function()
     vim.schedule(function()
       M.Draw(cancellationToken)
       vim.cmd("doautocmd User K8sDataLoaded")


Ramilito:
Hi!

Thank you for finding this issue and also for having a fix! Amazing!
I had forgotten about this use-case, sorry about that, we used to test this scenario but during the migration to use Rust it got lost.

I think this PR is OK to merge but there is one problem that would need to be taken care of in the future. With this PR merged, we would be creating new reflectors for every namespace per resource instead of for every resource. This could potentially cause the plugin to use more cpu/ram but I'm not sure on the numbers or if it's a problem yet.

There is a PR that aims to address it but for a different purpose #678

PR #737, opened 2026-01-28
feat: add nix flake

Ramilito:
add nix flake to support nix installations


Ramilito:
> why not use the make commands?

Those that run nix can't do that because nix assumes a sandboxed environment, so they don't run make build because that would require a global Go and Rust environment + download stuff from the internet. 
I'm no Nix pro but if just having a flake will enable them then lets add that.

PR #761, opened 2026-03-26
fix(helm): support Helm 4 CLI changes

tsacha:
## Summary
- **`helm list`**: drop the `-a` flag removed in Helm 4 (stderr was shown instead of JSON, breaking the view).
- **`helm status`**: drop `--show-desc`, removed in [helm/helm#13444](https://github.com/helm/helm/pull/13444); default status output still includes the release description.
- **`helm get manifest`**: pass `-n <namespace>` so the correct release is used when the same name exists in several namespaces.

## References
- [Helm changelog / overview](https://helm.sh/docs/changelog/) (Helm 4 list/status changes).

## Test plan
- [ ] Helm 4: open Helm view, describe (gd), YAML (gy), values (gv) on a release.
- [ ] Optional: Helm 3 still works with the same commands (flags dropped are redundant or default).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lua/kubectl/resources/helm/init.lua
+++ b/lua/kubectl/resources/helm/init.lua
@@ -11,7 +11,7 @@ local M = {
     display_name = "Helm",
     ft = "k8s_helm",
     cmd = "helm",
-    url = { "ls", "-a", "-A", "--output", "json" },
+    url = { "ls", "-A", "--output", "json" },
     hints = {
       { key = "<Plug>(kubectl.kill)", desc = "uninstall" },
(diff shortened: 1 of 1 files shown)


Ramilito:
Hey! Did something not work well? Do you need help with making helm v4 work?

tsacha:
Hello!

Indeed, I've been having some issues with Helm 4. I did a small test with a LLM and it created the PR a bit too quickly. :D The patch works well but I still need to verify backward compatibility with Helm v3.

PR #762, opened 2026-03-29
Helm 4 compatibility

tsacha:
## Summary
- **`helm list`**: drop the `-a` flag removed in Helm 4
- **`helm status`**: drop `--show-desc`, removed in [helm/helm#13444](https://github.com/helm/helm/pull/13444); default status output still includes the release description.
- **`helm get manifest`**: pass `-n <namespace>` so the correct release is used when the same name exists in several namespaces.

## Details

This PR adds Helm v4 compatibility and fixes a Lua error triggered when fetching releases:

```
Error executing vim.schedule lua callback: ...y/kubectl.nvim/lua/kubectl/resources/helm/definition.lua:10: bad argument #1 to 'ipairs' (table expected, got string)
stack traceback:
        [C]: in function 'ipairs'
        ...y/kubectl.nvim/lua/kubectl/resources/helm/definition.lua:10: in function 'processFunc'
        .../nvim/lazy/kubectl.nvim/lua/kubectl/resource_factory.lua:76: in function 'process'
        ...im/lazy/kubectl.nvim/lua/kubectl/resources/helm/init.lua:69: in function <...im/lazy/kubectl.nvim/lua/kubectl/resources/helm/init.lua:68>
```

Helm v4 deleted `-a` / `--all` flag from the `list` command:

```
$ helm version
version.BuildInfo{Version:"v4.1.3", GitCommit:"c94d381b03be117e7e57908edbf642104e00eb8f", GitTreeState:"", GoVersion:"go1.26.1-X:nodwarf5", KubeClientVersion:"v1.35"}
$ helm ls -a -A --output json
Error: unknown shorthand flag: 'a' in -a
```

An upstream PR ([helm/helm#31828](https://github.com/helm/helm/pull/31828)) aims to reintroduce it for backward compatibility, but it hasn't been merged yet.

*Note on breaking changes:*

This update changes how kubectl.nvim behaves with Helm v3: superseded, uninstalled, pending-install, and pending-upgrade releases will no longer be displayed in the list.

To maintain the previous behavior across both versions, we could use explicit flags: --pending --deployed --failed --superseded --uninstalled --uninstalling. Let me know if you'd like me to update the PR to maintain full backward compatibility.

### DIFF
--- a/lua/kubectl/resources/helm/init.lua
+++ b/lua/kubectl/resources/helm/init.lua
@@ -11,7 +11,7 @@ local M = {
     display_name = "Helm",
     ft = "k8s_helm",
     cmd = "helm",
-    url = { "ls", "-a", "-A", "--output", "json" },
+    url = { "ls", "-A", "--output", "json" },
     hints = {
       { key = "<Plug>(kubectl.kill)", desc = "uninstall" },
       { key = "<Plug>(kubectl.values)", desc = "values" },
@@ -43,7 +43,7 @@ end
 
 local function get_args()
   local ns_filter = state.getNamespace()
-  local args = add_namespace({ "ls", "-a", "--output", "json" }, ns_filter)
+  local args = add_namespace({ "ls", "--output", "json" }, ns_filter)
   return args
 end
 
--- a/lua/kubectl/resources/helm/init.lua
+++ b/lua/kubectl/resources/helm/init.lua
@@ -114,7 +114,7 @@ function M.Yaml(name, ns)
     display_name = M.definition.resource .. " | " .. name .. " | " .. ns,
     ft = "k8s_yaml",
     syntax = "yaml",
-    args = { "get", "manifest", name },
+    args = { "get", "manifest", name, "-n", ns },
   }
 
   local builder = manager.get_or_create(def.resource)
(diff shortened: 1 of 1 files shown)


Ramilito:
Hey! Awesome work!

I do want to stay backwards compatible with v3 somehow.
I suggest we do it in a simple manner by just adding a helm configuration option, what do you think? Alternatively we could dynamically figure out the version and store it when going to the helm view.

tsacha:
There you go. I tried the second solution with a sync shell command. The version is cached in M.version after the first call, so helm version is only executed once per session.

lua/kubectl/resources/helm/init.lua
```diff
@@ -114,7 +131,7 @@ function M.Yaml(name, ns)
     display_name = M.definition.resource .. " | " .. name .. " | " .. ns,
     ft = "k8s_yaml",
     syntax = "yaml",
-    args = { "get", "manifest", name },
+    args = { "get", "manifest", name, "-n", ns },
```
Ramilito, line 134: Don't we want the namespace here?

tsacha: We do, if we want to fetch YAML manifests from the All namespaces view.
We are getting the error `Error: release: not found` without this line.

Ramilito (approved):

Ramilito:
Amazing work! Thank you for sharing it 🙏

Issue #765, opened 2026-04-27
[BUG] Go cfgCache in DescribeResource poisoned when KUBECONFIG is unset (private EKS clusters via proxy)


AMEZIOUD-clever:
---                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
  ## Summary                                                                                                                             
                                                                                                                                       
  The `cfgCache` in `go/kubedescribe.go` is a process-lifetime `sync.Map` keyed by context name. When `KUBECONFIG` changes (or is        
  temporarily unset) between describe calls, the cached `rest.Config` becomes stale and points to the wrong server URL — specifically to
  the real cluster endpoint instead of a local proxy.                                                                                    
                                                                                                                                       
  ## Environment

  - kubectl.nvim v2.42.1                                                                                                                 
  - Kubernetes: EKS (private endpoint, eu-west-3)
  - Auth: credential helper (k8s-vault) generates a per-session kubeconfig with `server: https://127.0.0.1:<port>` (local proxy to the   
  private EKS API)                                                                                                                       
   
  ## Steps to reproduce                                                                                                                  
                                                                                                                                       
  1. Use a kubeconfig where `server` is a **local proxy** (e.g. `https://127.0.0.1:33290`). The real cluster endpoint is not reachable   
  from outside the VPC.
  2. `KUBECONFIG` is set to this proxy kubeconfig after the helper starts.                                                               
  3. Open kubectl.nvim, navigate to Nodes, run describe (`gd`) → works.                                                                
  4. Switch to another cluster: `KUBECONFIG` is briefly unset while the new proxy starts (~1 s gap).                                     
  5. During this gap, the describe session's 200 ms poll fires with `KUBECONFIG` unset.                                                  
  6. `getRestConfig("my-prod-eks-01")` → `NewDefaultClientConfigLoadingRules()` reads `~/.kube/config` (which has the real EKS        
  endpoint) → result is stored in `cfgCache`.                                                                                            
  7. Reconnect to the same cluster, describe again → Go uses the poisoned cache → tries to reach the real private endpoint → `i/o      
  timeout`.                                                                                                                              
                                                                                                                                       
  ## Error message                                                                                                                       
                                                                                                                                       
  ```
  Error describing nodes/ip-10-0-1-100.eu-west-3.compute.internal:
[...]

PR #766, opened 2026-04-30
fix(go): read KUBECONFIG via libc to honor host-side setenv

amezioud:
When kubectl.nvim is loaded as a cgo shared library inside an embedding process (e.g. Neovim), the Go runtime captures a snapshot of `environ` at startup. Subsequent setenv(3) calls performed by the host (e.g. vim.fn.setenv from Lua) update libc's environ but are invisible to os.Getenv on the Go side.

This causes KUBECONFIG changes made after the host has started -- typically when a plugin switches kube context after spawning a tunnel -- to be silently ignored. client-go falls back to ~/.kube/config and returns a *rest.Config pointing at the cluster's public URL instead of the user-provided kubeconfig (e.g. a forwarded local endpoint), leading to DNS/connection timeouts on private clusters.

Bypass the Go runtime cache by reading KUBECONFIG through C.getenv() and feeding it explicitly to clientcmd.LoadingRules.Precedence. The cache key in cfgCache also includes the path so that a KUBECONFIG switch invalidates the previously cached *rest.Config for the same context name.

Ref : https://github.com/Ramilito/kubectl.nvim/issues/765

### DIFF
--- a/go/kubedescribe.go
+++ b/go/kubedescribe.go
@@ -7,7 +7,9 @@ import "C"
 
 import (
 	"fmt"
+	"path/filepath"
 	"sync"
+	"unsafe"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


Ramilito:
Fixed in: #779

PR #767, opened 2026-05-28
feat: notify-on-binary-status

mosheavni:
Add some notifications for when the binary download has finished or has failed.

for example:

```
11:46:27 │ [kubectl.nvim] Binary not downloaded: ...pack/core/opt/blink.download/lua/blink/download/init.lua:37: No rust library found, but can't download due to not being on a git tag.. Ensure your plugin manager has kubectl.nvim checked out at a tagged release commit.
```

### DIFF
--- a/lua/kubectl/init.lua
+++ b/lua/kubectl/init.lua
@@ -244,7 +244,24 @@ function M.download_if_available(callback)
     root_dir = root_dir,
     output_dir = "/target/release",
     binary_name = "kubectl_client", -- excluding `lib` prefix
-  }, callback)
+  }, function(err)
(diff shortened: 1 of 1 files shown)


Ramilito (approved):

Ramilito:
Hey @mosheavni! Missed you!

Just checked that the blink.lib code only ever returns error, not sure how to make it safe for changes though so lets go with this!

mosheavni:
> Hey @mosheavni! Missed you!
> 
> Just checked that the blink.lib code only ever returns error, not sure how to make it safe for changes though so lets go with this!

glad to be back (:

PR #769, opened 2026-06-17
fix: detect changes when deleting fields on edit

mosheavni:
When editing a resource and only deleting lines, it notified "no changes detected" and never applied the changes.

The edit sent the whole object as a JSON merge patch, which can't represent deletions (an omitted key just means "leave unchanged"), so the dry-run result always matched the live object.

Now it computes a proper merge-patch diff between the live and edited object, so removed keys become explicit nulls. Deletions are detected and applied, while value changes keep working as before.


Ramilito:
Hey! Nice find, not sure how I've missed this for all this time...

Your fix seems to work but I would rather go in this direction #771, basically, let Kubernetes verify the change, WDYT?

PR #770, opened 2026-06-18
chore: don't add hints

mosheavni:
on yaml view for example, there are no hints. let's not add the line if there are no hints.


Ramilito:
The code looks awesome 👌.

Ramilito (approved):

PR #771, opened 2026-06-18
fix: edit now uses k8s to verify the changes

Ramilito:
Fixes: #769 

Moves responsibility to verify if something has changed or not to server side kubernetes. This should reduce the risk of us accidentally making an error.

Bonus:
Added more feedback to the user when things go wrong


mosheavni:
great! 🍇

PR #772, opened 2026-07-05
fix: edit async args

mosheavni:
Summary

Fixes resource edit (ge) failing on save/quit with an opaque userdata: 0x... notification instead of a real success or error message.

• Pass { path = tmpfilename } to edit_async so the Rust client receives {"path":"..."} instead of a JSON array
• Stringify errors (and non-string results) in run_async's vim.uv.new_work callback before returning to the main thread, so vim.notify shows the actual message

Problem

On :wq after editing a resource, users saw notifications like:

16:17:15 │ userdata: 0x01557783a0

The editor opened correctly (get_single_async uses the right payload shape), but the apply step never worked. edit_async expected CmdEditArgs { path } while Lua sent { tmpfilename }, which encodes as
["/tmp/..."]. Deserialization failed, and the mlua error lost its __tostring when crossing the worker thread boundary.

### DIFF
--- a/lua/kubectl/actions/commands.lua
+++ b/lua/kubectl/actions/commands.lua
@@ -245,7 +245,10 @@ function M.run_async(method_name, args, callback)
     local mod = require("kubectl_client")
     local ok, result = pcall(mod[method], json_str)
     if not ok then
-      return nil, result
+      return nil, tostring(result)
+    end
+    if type(result) ~= "string" then
+      return tostring(result), nil
     end
     return result, nil
   end
--- a/lua/kubectl/mappings.lua
+++ b/lua/kubectl/mappings.lua
@@ -274,7 +274,7 @@ function M.get_mappings()
               group = group,
               callback = function()
                 commands.run_async("edit_async", {
-                  tmpfilename,
+                  path = tmpfilename,
                 }, function(result, err)
                   vim.schedule(function()
                     if err then


Ramilito:
Hey!

This code seems weird to me, the userdata response means that it's a memory reference getting returned, doing a tostring on it won't resolve the memory ref since it's in a different thread. 

The solution should be to resolve the error on Rust side before returning it to the main lua thread.

Did you see it work differently?

mosheavni:
> Hey!
> 
> This code seems weird to me, the userdata response means that it's a memory reference getting returned, doing a tostring on it won't resolve the memory ref since it's in a different thread.
> 
> The solution should be to resolve the error on Rust side before returning it to the main lua thread.
> 
> Did you see it work differently?

I agree it seems weird, but for me it solved the issue. i was not able to edit HPA and after this I can

Ramilito:
Let me get home from vacation and take a look first, that piece of code handles all the async requests and need to be careful with it

mosheavni:
enjoy brother

lua/kubectl/actions/commands.lua
```diff
@@ -245,7 +245,10 @@ function M.run_async(method_name, args, callback)
     local mod = require("kubectl_client")
     local ok, result = pcall(mod[method], json_str)
     if not ok then
-      return nil, result
+      return nil, tostring(result)
```
Ramilito, line 248: Lets only do tostring on the error path and keep the other as is

lua/kubectl/mappings.lua
```diff
@@ -274,7 +274,7 @@ function M.get_mappings()
               group = group,
               callback = function()
                 commands.run_async("edit_async", {
-                  tmpfilename,
+                  path = tmpfilename,
```
Ramilito, line 277: this shouldn't be needed, lets revert that

Ramilito:
You were correct, this worked! Lets do some small changes though, added some comments

